PR #3, opened 2017-09-07
PP-2566 skeleton dropwizard 2

georgeracu:
This completes the Skeleton Dropwizard story. 
We added:
* db migration;
* guice;
* JPA;
* logging;
* refactored the app to rename it from pay-apps to pay-products
* updated README

The repository name will be changed soon.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,10 @@
 
 Products integrating with GOV.UK Pay in Java(Dropwizard)
 
+This microservice manages products (things that a Government Service would like to take payments for).
+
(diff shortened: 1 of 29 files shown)


README.md
```diff
@@ -1,12 +1,12 @@
-# pay-apps
+# pay-products
```
rauligs: repo called pay-apps and project pay-products might be confusing, not ideal

rauligs: Ups it is mentioned! all good

PR #8, opened 2017-09-28
PP-2668 adding create product endpoint

georgeracu:
*CONTEXT*
This PR introduces the very first API endpoint for products microservice. Create Product is expected to be called from `Selfservice` admin area in order to create a new sellable product (see product definition in ...) for a service. 
Note that given this is the first API endpoint we had to bring in few extra bits (test framework, etc) as well. Recommend to go by commits when reviewing. Below is a list of key changes 

*LIST OF CHANGES*
- Added catalogue and product entities and migrations scripts;
- Added testing infrastructure and created tests;
- Added DAOs for those entities;
- Added Product resource to handle POST requests to create a Product;
- Added Link, Errors and LinksBuilder for the response links;
- Added RequestValidator and created factory for Product;
- Updated README and definition document with new names;

*WHATS NOT DONE*
- OAuth based authentication. Will be done in a separate PR.


src/main/resources/config/config.yaml
```diff
@@ -59,4 +59,4 @@ jpa:
   cacheSharedDefault: false
 
 baseUrl: ${BASE_URL:-http://localhost:8080}
-frontendUrl: ${FRONTEND_URL:-http://localhost:3000/pay}
+productsUiPayUrl: ${PRODUCTSUI_PAY_URL:-http://localhost:3000/pay}
```
rauligs: - Minor. PRODUCTSUI or PRODUCTS_UI...
- Has this var been added to pay-ansible?

src/main/java/uk/gov/pay/products/service/ProductsCreator.java
```diff
@@ -0,0 +1,40 @@
...
+    private final LinksBuilder linksBuilder;
+
+    @Inject
+    public ProductsCreator(ProductDao productDao, LinksBuilder linksBuilder) {
+        this.productDao = productDao;
+        this.linksBuilder = linksBuilder;
+    }
+
+    @Transactional
+    public Product doCreate(JsonNode payload) {
```
rauligs: Not better ma json to domain before this method? and the signature would be _doCreate(product)_

src/main/java/uk/gov/pay/products/service/LinksBuilder.java
```diff
@@ -0,0 +1,37 @@
...
+import uk.gov.pay.products.model.Product;
+
+import java.net.URI;
+
+import static javax.ws.rs.HttpMethod.GET;
+import static javax.ws.rs.HttpMethod.POST;
+import static javax.ws.rs.core.UriBuilder.fromUri;
+import static uk.gov.pay.products.resources.ProductResource.PRODUCTS_RESOURCE;
+
+public class LinksBuilder {
```
rauligs: Calling this builder with a method decorate is not intuitive, maybe rename this to LinksDecorator? or something like that?

src/main/java/uk/gov/pay/products/validations/RequestValidations.java
```diff
@@ -0,0 +1,75 @@
...
+    public Optional<List<String>> checkMaxLength(JsonNode payload, int maxLength, String... fieldNames) {
+        return applyCheck(payload, exceedsMaxLength(maxLength), fieldNames, "Field [%s] must have a maximum length of " + maxLength + " characters");
+    }
+
+    private Function<JsonNode, Boolean> exceedsMaxLength(int maxLength) {
+        return jsonNode -> jsonNode.asText().length() > maxLength;
+    }
+
+    public Optional<List<String>> applyCheck(JsonNode payload, Function<JsonNode, Boolean> check, String[] fieldNames, String errorMessage) {
+        List<String> errors = newArrayList();
```
rauligs: This api is internal isn't? I would encourage failfast rather than validate all and aggregate them.

src/main/resources/config/config.yaml
georgeracu: This is already in the provision and deploy pipeline. I have to admit is a poor naming :(

src/main/java/uk/gov/pay/products/validations/RequestValidations.java
georgeracu: This is common functionality that we use in other microservices as well. The request validations can go into a commons package that we can pull from Maven and reuse it. I would rather keep it as it is and later on replace with the commons package.

rauligs: I meant changing it, but fine for now since that probably would take discussions with wider audience and agree to the approach.

PR #10, opened 2017-09-29
PP-2668 implemented OAuth2 based api authentication

kakumara:
All resource methods must now have a `@PermitAll` annotation and they will be automatically secured.

- Uses Dropwizard AuthDynamic feature with a custom authenticator.
- Added ability to support 2 configured auth api keys in order to support api token rotation. Made it safer to support both keys if and only if they are available, so we can choose to have only one configured.


src/test/java/uk/gov/pay/products/resources/ProductsResourceTest.java
```diff
@@ -122,7 +122,7 @@ public void shouldSuccess_whenSavingAValidProduct_withAllFields() throws Excepti
 
     @Test
     public void shouldError_whenSavingAProduct_withMandatoryFieldsMissing() throws Exception {
-        givenSetup()
+        givenAuthenticatedSetup()
```
georgeracu: I would also add a test to ProductResourceTest.java where no token is presented and should expect a 401.

src/main/java/uk/gov/pay/products/config/ProductsConfiguration.java
```diff
@@ -28,6 +28,9 @@
     @NotNull
     private String productsUiPayUrl;
 
+    private String apiToken;
```
georgeracu: A better naming like `authApiToken` will help, eg. `pay_api_token`

src/test/java/uk/gov/pay/products/resources/ProductsResourceTest.java
kakumara: There is already a test doing that 
https://github.com/alphagov/pay-products/pull/10/files#diff-adde91191a7a84bbc8b3967f3b3fd8deR14 
However, I also feel its better to have a test for each resource add specifically for 401

kakumara: Addressed 👍

PR #11, opened 2017-09-29
Pp 2669 existing product

georgeracu:
Get existing Product if exists, or return 404 otherwise.

With Authentication. 

Changes:
- Added `ProductFinder` and hooked into DI
- Added tests for finding an existing product and a non existing one;
- Added resource to handle GET requests to
`/v1/api/products/{externalProductId}`;
- Updated `ProductEntity` to avoid npe when creating a `Product` from an
incoming request that doesn't have a catalogue yet;
- Added `@Transactional` on finder to avoid stale data;
- Added methods on `DatabaseTestHelper` to persist a `Catalogue` and a
`Product` from tests;
- Added unit tests for `ProductFinder` and `ProductsCreator`;

Edit: added authentication, removed empty string validator, added unit tests;

### DIFF
--- a/src/main/java/uk/gov/pay/products/service/ProductFinder.java
+++ b/src/main/java/uk/gov/pay/products/service/ProductFinder.java
@@ -1,0 +1,23 @@
+package uk.gov.pay.products.service;
+
+import com.google.inject.Inject;
+import uk.gov.pay.products.model.Product;
+import uk.gov.pay.products.persistence.dao.ProductDao;
+
+import java.util.Optional;
+
+public class ProductFinder {
+    private final ProductDao productDao;
+    private final LinksDecorator linksDecorator;
+
+    @Inject
+    public ProductFinder(ProductDao productDao, LinksDecorator linksDecorator) {
+        this.productDao = productDao;
+        this.linksDecorator = linksDecorator;
+    }
+
+    public Optional<Product> findByExternalId(String externalId) {
+        return productDao.findByExternalId(externalId)
+                .map(productEntity -> linksDecorator.decorate(productEntity.toProduct()));
+    }
+}
--- a/src/main/java/uk/gov/pay/products/service/ProductFinder.java
+++ b/src/main/java/uk/gov/pay/products/service/ProductFinder.java
@@ -1,6 +1,7 @@
 package uk.gov.pay.products.service;
 
 import com.google.inject.Inject;
+import com.google.inject.persist.Transactional;
 import uk.gov.pay.products.model.Product;
 import uk.gov.pay.products.persistence.dao.ProductDao;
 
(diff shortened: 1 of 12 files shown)


src/main/java/uk/gov/pay/products/service/ProductFinder.java
```diff
@@ -0,0 +1,23 @@
...
+    private final ProductDao productDao;
+    private final LinksDecorator linksDecorator;
+
+    @Inject
+    public ProductFinder(ProductDao productDao, LinksDecorator linksDecorator) {
+        this.productDao = productDao;
+        this.linksDecorator = linksDecorator;
+    }
+
+    public Optional<Product> findByExternalId(String externalId) {
```
kakumara: Have a `@Transactional` annotation here, as discussed.

src/main/java/uk/gov/pay/products/validations/ProductRequestValidator.java
```diff
@@ -30,4 +32,11 @@ public ProductRequestValidator(RequestValidations requestValidations) {
 
         return errors.map(Errors::from);
     }
+
+    public Optional<Errors> validateFindRequest(String externalId) {
```
kakumara: I don't think it will reach here at all. Better to get rid of it

src/test/java/uk/gov/pay/products/resources/ProductsResourceTest.java
```diff
@@ -130,4 +137,56 @@ public void shouldError_whenSavingAProduct_withMandatoryFieldsMissing() throws E
                 .then()
                 .statusCode(400);
     }
+
+    @Test
+    public void givenAnExistingExternalProductId_shouldFindAndReturnProduct() throws Exception {
+        String externalId = randomUuid();
+        CatalogueEntity aCatalogueEntity = aCatalogueEntity().build();
+
+        ProductEntity product = ProductEntityFixture.aProductEntity()
```
kakumara: I would use `Product` here rather than `ProductEntity`

src/test/java/uk/gov/pay/products/utils/DatabaseTestHelper.java
```diff
@@ -10,4 +12,38 @@ public DatabaseTestHelper(DBI jdbi) {
         this.jdbi = jdbi;
     }
 
+    public DatabaseTestHelper addProduct(ProductEntity productEntity) {
```
kakumara: same as above .. use `Product`

src/main/java/uk/gov/pay/products/service/ProductFinder.java
```diff
@@ -0,0 +1,23 @@
...
+import uk.gov.pay.products.persistence.dao.ProductDao;
+
+import java.util.Optional;
+
+public class ProductFinder {
+    private final ProductDao productDao;
+    private final LinksDecorator linksDecorator;
+
+    @Inject
+    public ProductFinder(ProductDao productDao, LinksDecorator linksDecorator) {
```
kakumara: We need a unit test for these service class (mockito based) have a look at adminusers
Also i think ProductCreator missing unit tests as well

georgeracu:
PR comments have been addressed and added Authentication to the resource.

PR #13, opened 2017-10-03
PP-2670 Get a list of Products by externalServiceId

georgeracu:
- Added resource to GET a list of Product by a
query param externalServiceId;
- Added tests to test endpoint returning 200 and a list of existing
products, 404 when no products are found and 401 when not authenticated;
- Changed tests to allow adding Product and Catalogue in separate
operations;
- Updated LinksDecorator to decorate a list of products;

### DIFF
--- a/src/main/java/uk/gov/pay/products/persistence/dao/ProductDao.java
+++ b/src/main/java/uk/gov/pay/products/persistence/dao/ProductDao.java
@@ -4,6 +4,7 @@ import com.google.inject.Inject;
 import com.google.inject.Provider;
 import uk.gov.pay.products.model.Product;
 import uk.gov.pay.products.persistence.entity.ProductEntity;
+import uk.gov.pay.products.util.ProductStatus;
 
 import javax.persistence.EntityManager;
 import java.util.List;
@@ -29,11 +30,13 @@ public class ProductDao extends JpaDao<ProductEntity> {
 
     public List<Product> findByExternalServiceId(String externalServiceId) {
         String query = "SELECT product FROM ProductEntity product " +
-                "WHERE product.catalogueEntity.externalServiceId = :externalServiceId";
+                "WHERE product.catalogueEntity.externalServiceId = :externalServiceId " +
+                "AND product.status = :status";
 
         return entityManager.get()
                 .createQuery(query, ProductEntity.class)
                 .setParameter("externalServiceId", externalServiceId)
+                .setParameter("status", ProductStatus.ACTIVE)
                 .getResultList().stream()
                 .map(ProductEntity::toProduct)
                 .collect(Collectors.toList());
--- a/src/main/java/uk/gov/pay/products/persistence/dao/ProductDao.java
+++ b/src/main/java/uk/gov/pay/products/persistence/dao/ProductDao.java
@@ -2,10 +2,13 @@ package uk.gov.pay.products.persistence.dao;
 
 import com.google.inject.Inject;
 import com.google.inject.Provider;
+import uk.gov.pay.products.model.Product;
 import uk.gov.pay.products.persistence.entity.ProductEntity;
 
 import javax.persistence.EntityManager;
+import java.util.List;
 import java.util.Optional;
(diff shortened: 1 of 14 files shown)


src/main/java/uk/gov/pay/products/persistence/dao/ProductDao.java
```diff
@@ -23,4 +27,18 @@ protected ProductDao(Provider<EntityManager> entityManager) {
...
     }
+
+    public List<Product> findByExternalServiceId(String externalServiceId) {
+        String query = "SELECT product FROM ProductEntity product " +
+                "WHERE product.catalogueEntity.externalServiceId = :externalServiceId " +
+                "AND product.status = :status";
+
+        return entityManager.get()
+                .createQuery(query, ProductEntity.class)
+                .setParameter("externalServiceId", externalServiceId)
```
mrlumbu: index needed on externalServiceId?

mrlumbu: and maybe product status?

src/test/java/uk/gov/pay/products/resources/ProductsResourceTest.java
```diff
@@ -217,5 +217,60 @@ public void givenANotAuthenticatedRequest_shouldReturn401() throws Exception {
...
+                .then()
+                .statusCode(200);
+
+        response.body("", hasSize(2))
+                .body("[0].external_service_id", matchesPattern(externalServiceId))
+                .body("[0]._links", hasSize(2))
+                .body("[0].description", matchesPattern(product.getDescription()))
+                .body("[0].price", is(product.getPrice().intValue()))
+                .body("[0].name", matchesPattern(product.getName()))
+                .body("[1].name", matchesPattern(product_2.getName()));
```
mrlumbu: why are we not checking all things attributes of product 2?

src/test/java/uk/gov/pay/products/resources/ProductsResourceTest.java
```diff
@@ -217,5 +217,60 @@ public void givenANotAuthenticatedRequest_shouldReturn401() throws Exception {
...
+    }
+
+    @Test
+    public void givenNonExistingExternalServiceId_shouldReturn404() throws Exception {
+        givenAuthenticatedSetup()
+                .when()
+                .accept(APPLICATION_JSON)
+                .get(format("/v1/api/products?externalServiceId=%s", randomUuid()))
+                .then()
+                .statusCode(404);
```
mrlumbu: also need to test if an existing service with no active products returns 404?

src/test/java/uk/gov/pay/products/resources/ProductsResourceTest.java
georgeracu: Given that the above list is generated using a recursive call on a method that was tested earlier to generate the expected output I considered not testing for the second product as the functionality that will generate the values was tested above.

mrlumbu: fair enough 👍

src/test/java/uk/gov/pay/products/resources/ProductsResourceTest.java
georgeracu: :1st_place_medal:

src/main/resources/migrations/00001_add_catalogues_table.sql
```diff
@@ -12,3 +12,5 @@ CREATE TABLE catalogues (
 );
 --rollback drop table catalogues;
 
+--changeset uk.gov.pay:add_catalogues_external-service-id_index
+CREATE INDEX external_service_id_idx ON catalogues (external_service_id);
```
mrlumbu: I don't think this works - liquibase calculates checksums of already executed changes and stores them _somewhere_, so if you modify a changeset that's already been executed it'll complain and fail. I think you need to add a new changeset (`00003_blah`) with these changes. Worth running by Kasun, Raul or someone.

PR #14, opened 2017-10-04
PP-2671 Disable a Product

georgeracu:
- Product will be marked with status INACTIVE;
- Added status INACTIVE;
- Added test for ProductDao;
- Added new Resource with @DELETE;
- Added tests for resource;

### DIFF
--- a/src/main/java/uk/gov/pay/products/resources/ProductResource.java
+++ b/src/main/java/uk/gov/pay/products/resources/ProductResource.java
@@ -2,6 +2,7 @@ package uk.gov.pay.products.resources;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.google.inject.Inject;
+import io.dropwizard.jersey.PATCH;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.products.model.Product;
@@ -14,9 +15,7 @@ import javax.ws.rs.core.Response;
 import javax.ws.rs.core.Response.Status;
 
 import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
-import static javax.ws.rs.core.Response.Status.NOT_FOUND;
-import static javax.ws.rs.core.Response.Status.NO_CONTENT;
-import static javax.ws.rs.core.Response.Status.OK;
+import static javax.ws.rs.core.Response.Status.*;
 import static uk.gov.pay.products.resources.ProductResource.PRODUCTS_RESOURCE;
 
 @Path(PRODUCTS_RESOURCE)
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/pay/products/resources/ProductResource.java
```diff
@@ -62,4 +63,15 @@ public Response findProduct(@PathParam("productExternalId") String productExtern
                 .orElseGet(() ->
                         Response.status(NOT_FOUND).build());
     }
+
+    @DELETE
```
mrlumbu: Is there a reason why we consider this a delete, rather than disable. Not that it matters too much, but I would have said keeping the concept of disabling/enabling from resource down would be clearer, and would allow us to eg. introduce an 'enable' endpoint in future without confusing ourselves

src/test/java/uk/gov/pay/products/resources/ProductsResourceTest.java
```diff
@@ -217,5 +217,44 @@ public void givenANotAuthenticatedRequest_shouldReturn401() throws Exception {
...
+                .when()
+                .accept(APPLICATION_JSON)
+                .delete(format("/v1/api/products/%s", externalId))
+                .then()
+                .statusCode(204);
+
+    }
+
+    @Test
+    public void givenANonExistingExternalProductId_shouldReturn204() throws Exception{
```
mrlumbu: why 204 for non existent product?

src/main/java/uk/gov/pay/products/service/ProductFinder.java
```diff
@@ -22,4 +24,17 @@ public ProductFinder(ProductDao productDao, LinksDecorator linksDecorator) {
         return productDao.findByExternalId(externalId)
                 .map(productEntity -> linksDecorator.decorate(productEntity.toProduct()));
     }
+
+    @Transactional
+    public Optional<Product> disableProduct(String externalId) {
+        Optional<ProductEntity> productEntity = productDao.findByExternalId(externalId);
+        if(productEntity.isPresent()){
```
mrlumbu: sorry to come back on this again, but I think ideally this would follow our usual pattern of mapping over the optional returned by `findByExternalId`. It is good to have conventions 😄

src/main/java/uk/gov/pay/products/service/ProductFinder.java
```diff
@@ -22,4 +24,17 @@ public ProductFinder(ProductDao productDao, LinksDecorator linksDecorator) {
...
+            ProductEntity product = productEntity.get();
+            product.setStatus(ProductStatus.INACTIVE);
+            productDao.merge(product);
```
mrlumbu: don't think you need to merge here. In fact I think it could be harmful to.

PR #15, opened 2017-10-11
PP-2672 DB refactoring and code changes

SandorArpa:
## WHAT
DB structure changes to conform with the refined story. Products are no longer linked to a _Service_ but rather to a _Gateway_

## HOW

- Dropped table _catalogues_ as it is no longer fits the design
- Dropped column _catalogue_id_ from _products_ as it is no longer relevant
- Added column _gateway_account_id_ on table _products_ to provide linking to a _Gateway Account_
- Refactored code (removed CatalogueEntity, etc) to reflect DB changes
- Refactored tests to reflect code changes
- Updated documents (API specs and README)


src/main/resources/migrations/00005_drop_table_catalogues.sql
```diff
@@ -0,0 +1,7 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:drop_table_catalogues
+ALTER TABLE products DROP  CONSTRAINT fk_products_catalogues;
+
+DROP TABLE catalogues;
+
```
georgeracu: Don't we need rollback on this and 00006?

src/main/resources/migrations/00008_create_products_gateway_account_idx.sql
```diff
@@ -0,0 +1,6 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:add_index_to_gateway_account_id_on_products
+CREATE INDEX gateway_account_idx ON products(gateway_account_id);
+
+--rollback alter table products drop column gateway_account_id
```
georgeracu: rollback should probably drop the index and not the column

src/main/resources/migrations/00005_drop_table_catalogues.sql
SandorArpa: We can't really roll back the drop of a table or a column. Either the table (or column) been dropped, in which case we lost the existing data or the drop command hasn't been executed, in which case the table (or column) is retained with the existing data.

src/main/resources/migrations/00008_create_products_gateway_account_idx.sql
SandorArpa: Spot on, thank you for noticing it!

src/main/resources/migrations/00005_drop_table_catalogues.sql
georgeracu: Thanks :)

PR #22, opened 2017-11-01
PP-2672 Renamed `productUiUrl` to `productUiPayUrl`

ckalista:
#WHAT 
This config name change is to make it more explicit that the URL expected here is not the Product UI base URL but the full URL to the paying resource.


src/main/java/uk/gov/pay/products/config/ProductsConfiguration.java
```diff
@@ -29,7 +29,7 @@
     private String publicApiUrl;
 
     @NotNull
-    private String productsUiUrl;
+    private String productsUiPayUrl;
```
kakumara: 😍  it was named like this before for a reason 👍

PR #23, opened 2017-11-02
PP-2672 Refactor GET payments API

SandorArpa:
## WHAT
Refactor GET payments API

## HOW

- refactor ProductEntity to link PaymentEntity by LAZY loading
- refactor PaymentFinder to return a list of Payments through ProductDao
- remove findProductIdByExternalId from ProductDao
- refactor tests accordingly


src/main/java/uk/gov/pay/products/persistence/entity/ProductEntity.java
```diff
@@ -45,6 +49,9 @@
     @Column(name = "return_url")
     private String returnUrl;
 
+    @OneToMany(mappedBy = "productEntity", targetEntity = PaymentEntity.class, fetch = FetchType.LAZY)
+    private List<PaymentEntity> payments = new ArrayList<>();
```
kakumara: I'm bit concerned about mapping payments collection (even if lazy) in product object. I'd rather not do it and instead use a JOIN JPA query to list the payments for a product. 
Imagine if the product is something like  `pay for Tax disc` or something popular. We are talking about millions of payments, which (even if lazy) will kill the application if we were trying load.

In the Join option we always have a choice to LIMIT results and even  support pagination at some point (i'm pretty sure we going to need it at some point)

kakumara: I'd rather have a opposite side relationship. A `@ManyToOne` from PaymentEntity to ProductEntity. That'll help the JOIN query too.

src/main/java/uk/gov/pay/products/persistence/entity/ProductEntity.java
```diff
@@ -8,9 +8,13 @@
 import javax.persistence.Entity;
 import javax.persistence.EnumType;
 import javax.persistence.Enumerated;
+import javax.persistence.FetchType;
+import javax.persistence.OneToMany;
```
kakumara: surely and unused import ???

src/test/java/uk/gov/pay/products/service/PaymentFinderTest.java
```diff
@@ -26,6 +27,9 @@
     @Mock
     private PaymentDao paymentDao;
 
+    @Mock
+    private ProductEntity mockProductEntity;
```
kakumara: Is this used now ?

PR #24, opened 2017-11-02
PP-2672 Improved version of `PublicApiRestClient` taking into account the `Bearer` API token

ckalista:
# WHAT

Improved version of `PublicApiRestClient` taking into account the `Bearer` API token that needs to be passed in the header of all PublicAPI requests for obvious security reasons. The authorization token is in the form of `Bearer + API token` where the API token is specific for each product.

### DIFF
--- a/src/test/java/uk/gov/pay/products/client/publicapi/PublicApiRestClientTest.java
+++ b/src/test/java/uk/gov/pay/products/client/publicapi/PublicApiRestClientTest.java
@@ -58,7 +58,7 @@ public class PublicApiRestClientTest {
     }
 
     @Test
-    public void createPayment_shouldThrowAnException() throws PublicApiResponseErrorException {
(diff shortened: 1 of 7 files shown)


src/test/java/uk/gov/pay/products/client/publicapi/PublicApiRestClientTest.java
```diff
@@ -128,7 +131,7 @@ public void findPayment_shouldThrowAnException() {
                 .respondBadRequestWithBody(errorPayload);
 
         try {
-            publicApiRestClient.getPayment(paymentId);
+            publicApiRestClient.getPayment(apiToken, paymentId);
```
kakumara: Hoe about a test for getting 401 from PublicApi for an invalid api token?

ckalista: Yes, I have added the test. 👍

PR #29, opened 2017-11-07
PP 2672 Check if return url is valid url

SandorArpa:
- Update RequestValidations to check whether the return URL is a valid url
- Make return_url non compulsory
- Add handling of return url. When return url is not present (prototype case) then we add a pre-defined return url
- Add/refactor tests accordingly

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -19,6 +19,7 @@
         <guice.version>4.1.0</guice.version>
         <hamcrest.version>1.3</hamcrest.version>
         <mainClass>uk.gov.pay.products.ProductsApplication</mainClass>
+        <powermock.version>1.7.3</powermock.version>
     </properties>
 
     <dependencyManagement>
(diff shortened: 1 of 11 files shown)


kakumara:
There is a problem with the commits / rebasing. I can see a merge commit in the commits. Please fix it.

pom.xml
```diff
@@ -175,6 +176,18 @@
             <version>3.4.6</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.powermock</groupId>
```
kakumara: What's the motive behind using powermock? So far we'ven been okay with using just `mockito`

src/main/java/uk/gov/pay/products/config/ProductsConfiguration.java
```diff
@@ -32,8 +32,12 @@
     private String productsUiPayUrl;
 
     private String productsApiToken;
+
     private String productsApiToken2;
 
+    @NotNull
+    private String productsUiConfirmUrl;
```
kakumara: Are you sure, is this part of this PR?

SandorArpa:
there was a mix up and had to cherry pick into a new one

PR #30, opened 2017-11-07
Pp 2672 Check valid  and handle return url

SandorArpa:
- Update RequestValidations to check whether the return URL is a valid url
- Make return_url non compulsory
- Add handling of return url. When return url is not present (prototype case) then we add a pre-defined return url

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -19,6 +19,7 @@
         <guice.version>4.1.0</guice.version>
         <hamcrest.version>1.3</hamcrest.version>
         <mainClass>uk.gov.pay.products.ProductsApplication</mainClass>
+        <powermock.version>1.7.3</powermock.version>
     </properties>
 
     <dependencyManagement>
@@ -154,7 +155,7 @@
         <dependency>
             <groupId>org.mockito</groupId>
             <artifactId>mockito-core</artifactId>
-            <version>2.10.0</version>
+            <version>2.8.47</version>
             <scope>test</scope>
         </dependency>
         <dependency>
@@ -173,6 +174,18 @@
             <groupId>org.unitils</groupId>
             <artifactId>unitils-core</artifactId>
             <version>3.4.6</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.powermock</groupId>
+            <artifactId>powermock-module-junit4</artifactId>
+            <version>${powermock.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.powermock</groupId>
+            <artifactId>powermock-api-mockito2</artifactId>
+            <version>${powermock.version}</version>
             <scope>test</scope>
         </dependency>
     </dependencies>
--- a/src/main/java/uk/gov/pay/products/validations/ProductRequestValidator.java
+++ b/src/main/java/uk/gov/pay/products/validations/ProductRequestValidator.java
@@ -13,7 +13,7 @@ public class ProductRequestValidator {
     private static final String FIELD_PAY_API_TOKEN = "pay_api_token";
     private static final String FIELD_NAME = "name";
     private static final String FIELD_PRICE = "price";
-    private static final String RETURN_URL = "return_url";
+    private static final String FIELD_RETURN_URL = "return_url";
 
 
     @Inject
@@ -28,7 +28,11 @@ public class ProductRequestValidator {
                 FIELD_PAY_API_TOKEN,
                 FIELD_NAME,
                 FIELD_PRICE,
-                RETURN_URL);
+                FIELD_RETURN_URL);
+
+        if(!errors.isPresent()){
+            errors = requestValidations.checkIsUrl(payload, FIELD_RETURN_URL);
+        }
 
         return errors.map(Errors::from);
     }
--- a/src/main/java/uk/gov/pay/products/validations/ProductRequestValidator.java
+++ b/src/main/java/uk/gov/pay/products/validations/ProductRequestValidator.java
@@ -30,7 +30,7 @@ public class ProductRequestValidator {
                 FIELD_PRICE);
 
         if(!errors.isPresent()){
-            if(!requestValidations.checkIfExists(payload, FIELD_RETURN_URL).isPresent()) {
+            if( payload.get(FIELD_RETURN_URL)!= null ) {
                 errors = requestValidations.checkIsUrl(payload, FIELD_RETURN_URL);
             }
         }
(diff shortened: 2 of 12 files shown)


pom.xml
```diff
@@ -19,6 +19,7 @@
         <guice.version>4.1.0</guice.version>
         <hamcrest.version>1.3</hamcrest.version>
         <mainClass>uk.gov.pay.products.ProductsApplication</mainClass>
+        <powermock.version>1.7.3</powermock.version>
```
kakumara: I would still question the use of powermock/static mocking. This is usually discouraged and there is normally an alternative. We have been using `randomUuid()` elsewhere (adminusers) heavily and we never felt the need to have powermock for testing there. Mockito was good enough

pom.xml
```diff
@@ -154,7 +155,7 @@
         <dependency>
             <groupId>org.mockito</groupId>
             <artifactId>mockito-core</artifactId>
-            <version>2.10.0</version>
+            <version>2.8.47</version>
```
kakumara: Any reason for the downgrade?

src/main/java/uk/gov/pay/products/validations/ProductRequestValidator.java
```diff
@@ -27,8 +27,13 @@ public ProductRequestValidator(RequestValidations requestValidations) {
                 FIELD_GATEWAY_ACCOUNT_ID,
                 FIELD_PAY_API_TOKEN,
                 FIELD_NAME,
-                FIELD_PRICE,
-                RETURN_URL);
+                FIELD_PRICE);
+
+        if(!errors.isPresent()){
+            if(!requestValidations.checkIfExists(payload, FIELD_RETURN_URL).isPresent()) {
```
kakumara: 😖 seriously ? Do you really need requestValidations to check if the `payload` having a field named `FIELD_RETUEN_URL`?

src/main/java/uk/gov/pay/products/validations/ProductRequestValidator.java
```diff
@@ -27,8 +27,13 @@ public ProductRequestValidator(RequestValidations requestValidations) {
...
+                errors = requestValidations.checkIsUrl(payload, FIELD_RETURN_URL);
+            }
```
kakumara: Don't we have to check if the `return_url` is `https` too? or are we doing it somewhere else?

src/main/resources/config/config.yaml
```diff
@@ -64,5 +64,5 @@ jerseyClientConfiguration:
 baseUrl: ${BASE_URL}
 publicApiUrl: ${PUBLICAPI_URL}
 productsUiPayUrl: ${PRODUCTSUI_PAY_URL}
-
+productsUiConfirmUrl: ${PRODUCTSUI_CONFIRMATION_URL}
```
kakumara: suggestion: Normally I would put all the local urls as default values for all these URL configs if environment variable is not present. So that we don't have to keep maintaining `pay-scripts`. 
For e.g. 
```
productsUiPayUrl: ${PRODUCTSUI_PAY_URL:-https://productsui.pymnt.localdomain/pay}
```

src/test/java/uk/gov/pay/products/service/PaymentCreatorTest.java
```diff
@@ -86,10 +95,76 @@ public void shouldCreateASuccessfulPayment() throws Exception {
         PaymentResponse paymentResponse = createPaymentResponse(
                 paymentId,
                 paymentAmount,
-                paymentNextUrl);
+                paymentNextUrl,
+                productReturnUrl);
+
+
+        when(productDao.findByExternalId(productExternalId)).thenReturn(Optional.of(productEntity));
+        when(RandomIdGenerator.randomUuid()).thenReturn(paymentExternalId);
```
kakumara: For this test you are not even using this randomUuid() for anything. So why bother power mocking?

src/test/java/uk/gov/pay/products/service/PaymentCreatorTest.java
```diff
@@ -86,10 +95,76 @@ public void shouldCreateASuccessfulPayment() throws Exception {
...
+                productPrice,
+                productExternalId,
+                productName,
+                "",
+                productApiToken);
+        PaymentRequest expectedPaymentRequest = createPaymentRequest(
+                productPrice,
+                productExternalId,
+                productName,
+                productReturnUrl + "/" + paymentExternalId);
```
kakumara: For this test it makes slight sense. But, just a pattern match would be sufficient than a exact match in these cases.

pom.xml
SandorArpa: Powermock compatibility

src/main/java/uk/gov/pay/products/validations/ProductRequestValidator.java
SandorArpa: Validation is on front-end but can add the extra check

src/test/java/uk/gov/pay/products/service/PaymentCreatorTest.java
SandorArpa: Because we mock `RandomIdGenerator` in `setup()` we need to tell it what to return when running the test otherwise `PaymentCreator.beforePaymentCreation(...)` fails to assign an external id. I can move it to the mock within the test method I assume.

src/test/java/uk/gov/pay/products/service/PaymentCreatorTest.java
SandorArpa: It is because how `PaymentRequestChecker` is written. `isSame()` would fail if the two external ids weren't the same.

pom.xml
SandorArpa: That's true. On the other hand `adminusers` is not using the complex `TransactionFlow` process.

src/main/java/uk/gov/pay/products/validations/ProductRequestValidator.java
SandorArpa: 👍

src/main/resources/config/config.yaml
SandorArpa: This can be a separate PR as none of the previous env vars were set this way and thus it is a 'feature' request.

pom.xml
kakumara: Disagree. TransactionFlow has nothing to do with the usage of static contexts.

pom.xml
kakumara: 😖

src/main/resources/config/config.yaml
kakumara: its fine. Not the end of the world!

kakumara:
I'd given up on powermock comments, although I disagree to some of the responses. Given its only a test dependency.

src/main/resources/config/config.yaml
SandorArpa: I agree with your comment and I shall do it in the next PR for all the vars.

src/main/java/uk/gov/pay/products/validations/ProductRequestValidator.java
SandorArpa: Fair enough, I shall change it 👍

PR #46, opened 2017-12-08
PP-3031 Random Reference Number DB changes

SandorArpa:
- add column gateway_account_id to payments
- add column reference_number to payments

### DIFF
--- a/src/main/resources/migrations/00011_add_column_gateway_account_id_to_payments.sql
+++ b/src/main/resources/migrations/00011_add_column_gateway_account_id_to_payments.sql
@@ -1,0 +1,6 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:add_column_gateway_account_id_on_payments
+ALTER TABLE payments ADD COLUMN gateway_account_id BIGINT;
+
+--rollback alter table payments drop column gateway_account_id;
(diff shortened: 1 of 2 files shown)


kakumara:
I think there's few things missing from this PR. 
we need to assume there is already records in payments table. and we cannot have nulls in these columns and must be unique.
So my approach for  this would be;
1. add 2 columns (nuallable) in to the payments table.
2. write migrate scripts to copy corresponding gateway_account_id column from `products` table to `payments` table
3. Write migration script to generate random 10 char value to reference column. This does not have to follow the new algorithm. I would just use `uuid_generate_v4()` function and chop `-` off and then take  first 10 characters.
4. Finally make the combination of two columns unique.

src/main/resources/migrations/00014_update_non_existestent_reference_ids_in_payments.sql
```diff
@@ -0,0 +1,5 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:update_non_existestent_reference_ids_in_payments
+
+UPDATE payments SET reference_number = upper(substring(replace(uuid_generate_v4()::VARCHAR,'-',''), 1, 10));
```
kakumara: Nice 👍

PR #55, opened 2017-12-18
PP-3035 add service name

maxcbc:
- Adds `service_name` column to product in DB (with no constraints for the time being)
- Adds `service_name` to product
- Adds `service_name` as required field when creating a product

### DIFF
--- a/src/main/resources/migrations/00017_add_column_service_name_to_products.sql
+++ b/src/main/resources/migrations/00017_add_column_service_name_to_products.sql
@@ -1,0 +1,6 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:add_column_service_name_on_products
+ALTER TABLE products ADD COLUMN service_name VARCHAR(255);
+
+--rollback alter table products drop column service_name;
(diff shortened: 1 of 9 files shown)


src/main/resources/migrations/00017_add_column_service_name_to_products.sql
```diff
@@ -0,0 +1,6 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:add_column_service_name_on_products
+ALTER TABLE products ADD COLUMN service_name VARCHAR(255);
```
kakumara: This creates a `Nullable` column. I think this should be 'NOT NULL' 
And for the records that are already in perhaps we should just make it `default 'System Generated'`

maxcbc: I disagree, I think it should be `Nullable` for now. If it comes back to pay-products as blank, it shows nothing. If we default it to `System Generated` it will show `System Generated` in the top bar. Once we move this into AWS, all newly created products will have a value, there will be no `Null` values so we can add the constraint then.

kakumara:
Assuming we will put the constraints when we are in AWS

maxcbc:
@kakumara See PP-3086

PR #57, opened 2018-01-11
PP-3048 Keep service name in sync

SandorArpa:
Implement GatewayAccountResource to update Service Name in `products`

- add validator [ece1427](https://github.com/alphagov/pay-products/commit/ece1427847646fecc58a208bae5d80287fac53de)
- add GatewayAccountUpdater and GatewayAccountRequest, refactor Product to create Product from ProductEntity [5b4f6dc](https://github.com/alphagov/pay-products/commit/5b4f6dc48a8ae8305bfcbd0dd9ba7defc78bc95a)
- add GatewayAccountFactory and configure ProductsApplication to inject GatewayAccountFactory [e5b061b](https://github.com/alphagov/pay-products/commit/e5b061b89c15aed10179c48477b59f19a54dc77d)
- add GatewayAccountResource and configuration to Application and Module [ecd5ea3](https://github.com/alphagov/pay-products/commit/ecd5ea34f36a57964d594db18b5540b639792ab9)
- add/refactor relevant tests
- address PR comments [47dc1d8](https://github.com/alphagov/pay-products/pull/57/commits/47dc1d8c1fc66f7fde1eeeec77046b40d01fda55)

### DIFF
--- a/src/main/java/uk/gov/pay/products/model/GatewayAccountRequest.java
+++ b/src/main/java/uk/gov/pay/products/model/GatewayAccountRequest.java
@@ -1,0 +1,78 @@
+package uk.gov.pay.products.model;
+
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+
+import static com.google.common.collect.Lists.newArrayList;
+import static java.lang.String.format;
+import static java.util.stream.Collectors.toList;
+import static org.apache.commons.lang3.StringUtils.isEmpty;
+
+public class GatewayAccountRequest {
+
+    public static final String FIELD_OPERATION = "op";
+    public static final String FIELD_OPERATION_PATH = "path";
+    public static final String FIELD_VALUE = "value";
+
+    private String op;
+    private String path;
+    private JsonNode value;
+
+    public String getOp() {
+        return op;
+    }
+
+    public String getPath() {
+        return path;
+    }
+
+    public String valueAsString() {
+        if (value != null && value.isTextual()) {
+            return value.asText();
+        }
+        return null;
+    }
+
+    public List<String> valueAsList() {
+        if (value != null && value.isArray()) {
+            return newArrayList(value.elements())
+                    .stream()
+                    .map(node -> node.textValue())
+                    .collect(toList());
+        }
+        return null;
+    }
+
+    public Map<String, String> valueAsObject() {
+        if (value != null) {
+            if ((value.isTextual() && !isEmpty(value.asText())) || (!value.isNull() && value.isObject())) {
+                try {
+                    return new ObjectMapper().readValue(value.traverse(), new TypeReference<Map<String, String>>() {});
+                } catch (IOException e) {
+                    throw new RuntimeException(format("Malformed JSON object in GatewayAccountRequest.value"), e);
+                }
+            }
+        }
+        return null;
+    }
+
+
+    private GatewayAccountRequest(String op, String path, JsonNode value) {
+        this.op = op;
+        this.path = path;
+        this.value = value;
+    }
+
+    public static GatewayAccountRequest from(JsonNode payload) {
+        return new GatewayAccountRequest(
+                payload.get(FIELD_OPERATION).asText(),
+                payload.get(FIELD_OPERATION_PATH).asText(),
+                payload.get(FIELD_VALUE));
+
+    }
+}
--- a/src/main/java/uk/gov/pay/products/model/Product.java
+++ b/src/main/java/uk/gov/pay/products/model/Product.java
@@ -6,6 +6,7 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.PropertyNamingStrategy;
 import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import uk.gov.pay.products.persistence.entity.ProductEntity;
 import uk.gov.pay.products.util.ProductStatus;
 import uk.gov.pay.products.util.ProductType;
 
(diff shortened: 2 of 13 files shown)


src/main/java/uk/gov/pay/products/model/GatewayAccountRequest.java
```diff
@@ -0,0 +1,78 @@
...
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+
+import static com.google.common.collect.Lists.newArrayList;
+import static java.lang.String.format;
+import static java.util.stream.Collectors.toList;
+import static org.apache.commons.lang3.StringUtils.isEmpty;
+
+public class GatewayAccountRequest {
```
kakumara: These PATCH requests classes look identical. They all have the same 3 fields and same sort of operations. Wonder if we can generify them so that any future PATCH'es will become much more simpler

src/main/java/uk/gov/pay/products/model/Product.java
```diff
@@ -79,6 +80,12 @@ public static Product from(JsonNode jsonPayload) {
                 price, ProductStatus.ACTIVE, gatewayAccountId, serviceName, type, returnUrl);
     }
 
+    public static Product valueOf(ProductEntity productEntity) {
```
kakumara: 👎  you should do this in the ProductEntity. Not in Product. The reason being, Product should never need to know about Entity classes and should avoid any dependencies being loaded. 

On the other hand its Okay for ProductEntity to know about Product (so the dependency)

src/main/java/uk/gov/pay/products/resources/GatewayAccountResource.java
```diff
@@ -0,0 +1,61 @@
...
+import static javax.ws.rs.core.Response.Status.NOT_FOUND;
+
+@Path("/")
+public class GatewayAccountResource {
+
+    private static Logger logger = LoggerFactory.getLogger(GatewayAccountResource.class);
+
+    private static final String API_VERSION_PATH = "v1";
+    public static final String GATEWAY_ACCOUNT_RESOURCE_PATH = API_VERSION_PATH + "/api/gateway-account";
+    public static final String GATEWAY_ACCOUNT_API_PATH = GATEWAY_ACCOUNT_RESOURCE_PATH + "/{gatewayAccountId}";
```
kakumara: how about `GATEWAY_ACCOUNTS_PATH` for `/api/gateway-account`
and `GATEWAY_ACCOUNT_PATH` for `/api/gateway-account/{gatewayAccountId}`
coz they gives singular and plural meaning correctly

src/main/java/uk/gov/pay/products/resources/GatewayAccountResource.java
```diff
@@ -0,0 +1,61 @@
...
+    @Path(GATEWAY_ACCOUNT_API_PATH)
+    @Consumes(APPLICATION_JSON)
+    @Produces(APPLICATION_JSON)
+    @PermitAll
+    public Response patchGatewayAccount(@PathParam("gatewayAccountId") Integer gatewayAccountId, JsonNode payload) {
+        logger.info("Patching gateway account [ {} ]", gatewayAccountId);
+        return requestValidator.validatePatchRequest(payload)
+                .map(errors -> Response.status(BAD_REQUEST).entity(errors).build())
+                .orElseGet(() -> {
+                            List<Product> products = gatewayAccountFactory.getUpdateService()
```
kakumara: Why List<Product> returned here? Is that necessary? There could be hundreds (or thousands) of Products just getting loaded for no reason?
Ideally it should be Optional<GatewayAccount> but since we don't have that entity perhaps we can just stick with a `Boolean`

src/main/java/uk/gov/pay/products/service/GatewayAccountUpdater.java
```diff
@@ -0,0 +1,36 @@
...
+    private final ProductDao productDao;
+
+    @Inject
+    public GatewayAccountUpdater(ProductDao productDao) {
+        this.productDao = productDao;
+    }
+
+    @Transactional
+    public List<Product> doPatch(Integer gatewayAccountId, GatewayAccountRequest gatewayAccountRequest) {
+        List<ProductEntity> productEntities = productDao.findByGatewayAccountId(gatewayAccountId);
```
kakumara: Rather than loading all entities and looping update queries can't we think of a one update query affecting all rows without loading anything? That'd be more efficient

src/main/java/uk/gov/pay/products/validations/GatewayAccountRequestValidator.java
```diff
@@ -0,0 +1,32 @@
...
+    @Inject
+    public GatewayAccountRequestValidator(RequestValidations requestValidations) {
+        this.requestValidations = requestValidations;
+    }
+
+    public Optional<Errors> validatePatchRequest(JsonNode payload) {
+        Optional<List<String>> errors = requestValidations.checkIfExists(
+                payload,
+                FIELD_OPERATION,
+                FIELD_OPERATION_PATH,
```
kakumara: we only support `service-name`. So perhaps we need to validate value in op_path too.

src/main/java/uk/gov/pay/products/resources/GatewayAccountResource.java
SandorArpa: 👍

src/main/java/uk/gov/pay/products/model/GatewayAccountRequest.java
SandorArpa: 👍

src/main/java/uk/gov/pay/products/model/Product.java
SandorArpa: It is no longer needed as GatewayAccountUpdater refactored to return a Boolean instead of a List of Products

src/main/java/uk/gov/pay/products/resources/GatewayAccountResource.java
SandorArpa: 👍

src/main/java/uk/gov/pay/products/validations/GatewayAccountRequestValidator.java
SandorArpa: That's coming in part 2 (using a BiConsumer)

src/main/java/uk/gov/pay/products/service/GatewayAccountUpdater.java
SandorArpa: Will address it in part 2

PR #59, opened 2018-01-15
PP-3034 payment price override

kakumara:
As part of adhoc payment journey, the paying user is able to provide the payment amount. This requires using that captured amount used during payment/charge creation which actually overrides the products price. 
There is not special CAP at the moment and the user is able to provide the maximum amount that payment platform can take at the moment.

### DIFF
--- a/src/main/java/uk/gov/pay/products/resources/PaymentResource.java
+++ b/src/main/java/uk/gov/pay/products/resources/PaymentResource.java
@@ -1,11 +1,14 @@
 package uk.gov.pay.products.resources;
 
+import com.fasterxml.jackson.databind.JsonNode;
 import com.google.inject.Inject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.products.model.Payment;
 import uk.gov.pay.products.service.PaymentFactory;
 import uk.gov.pay.products.service.ProductFactory;
+import uk.gov.pay.products.validations.PaymentRequestValidator;
+import uk.gov.pay.products.validations.ProductRequestValidator;
 
 import javax.annotation.security.PermitAll;
 import javax.ws.rs.Consumes;
@@ -34,12 +37,12 @@ public class PaymentResource {
     public static final String PRODUCT_PAYMENTS_RESOURCE_PATH = PRODUCT_RESOURCE_PATH + "/payments";
 
     private final PaymentFactory paymentFactory;
-    private final ProductFactory productFactory;
+    private final PaymentRequestValidator requestValidator;
 
     @Inject
-    public PaymentResource(PaymentFactory paymentFactory, ProductFactory productFactory) {
+    public PaymentResource(PaymentFactory paymentFactory, PaymentRequestValidator requestValidator) {
         this.paymentFactory = paymentFactory;
-        this.productFactory = productFactory;
+        this.requestValidator = requestValidator;
     }
 
     @Path(PAYMENT_RESOURCE_PATH)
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/pay/products/resources/PaymentResource.java
```diff
@@ -61,10 +64,22 @@ public Response findPaymentByExternalId(@PathParam("paymentExternalId") String p
...
-    public Response createPaymentByProductExternalId(@PathParam("productExternalId") String productExternalId) {
+    public Response createPayment(@PathParam("productExternalId") String productExternalId, JsonNode priceOverride) {
         logger.info("Create a payment for product id - [ {} ]", productExternalId);
-        Payment payment = paymentFactory.paymentCreator().doCreate(productExternalId);
-        return Response.status(CREATED).entity(payment).build();
+        return requestValidator.validatePriceOverrideRequest(priceOverride)
+                .map(errors -> Response.status(Response.Status.BAD_REQUEST).entity(errors).build())
+                .orElseGet(() -> {
+                    Payment payment = paymentFactory.paymentCreator().doCreate(productExternalId, extractAmountIfAvailable(priceOverride));
+                    return Response.status(CREATED).entity(payment).build();
```
SandorArpa: What if Payment Creation fails?

src/test/java/uk/gov/pay/products/resources/PaymentResourceTest.java
```diff
@@ -17,9 +18,7 @@
 
 import static java.lang.String.format;
 import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
-import static org.hamcrest.Matchers.hasEntry;
-import static org.hamcrest.Matchers.hasKey;
-import static org.hamcrest.Matchers.hasSize;
+import static org.hamcrest.Matchers.*;
```
SandorArpa: grrrrr

src/test/java/uk/gov/pay/products/service/PaymentCreatorTest.java
```diff
@@ -206,6 +206,64 @@ public void shouldCreateASuccessfulPayment_whenReturnUrlIsNotPresent() throws Ex
         verify(paymentDao).merge(argThat(PaymentEntityMatcher.isSame(expectedPaymentEntity)));
     }
 
+    @Test
+    public void shouldCreateASuccessfulPayment_withOverridePrice_whenPriceOverridePresent() throws Exception {
+        PowerMockito.mockStatic(RandomIdGenerator.class);
+        int productId = 1;
+        String productExternalId = "product-external-id";
```
SandorArpa: Shouldn't we rather use randomUID() for any/all externalIds?

src/test/java/uk/gov/pay/products/validations/PaymentRequestValidatorTest.java
```diff
@@ -0,0 +1,71 @@
...
+
+    @Before
+    public void before() throws Exception {
+        requestValidator = new PaymentRequestValidator(new RequestValidations());
+    }
+
+    @Test
+    public void shouldSuccess_onCreatePayment_ifJsonPayloadIsNull() throws Exception {
+        Optional<Errors> errors = requestValidator.validatePriceOverrideRequest(null);
+        assertFalse(errors.isPresent());
```
SandorArpa: assertThat(..., is()); perhaps?

src/test/java/uk/gov/pay/products/resources/PaymentResourceTest.java
kakumara: Whats wrong with that. In fact its intelliJ magic

src/main/java/uk/gov/pay/products/resources/PaymentResource.java
kakumara: I believe it will be an 500. I didn't change the existing logic here. Only wrapped a validation around the existing logic

src/test/java/uk/gov/pay/products/resources/PaymentResourceTest.java
SandorArpa: Well, we had an agreement with the team sometime ago that we don't use `import xxx.*`

src/test/java/uk/gov/pay/products/validations/PaymentRequestValidatorTest.java
kakumara: not sure we actually standardise on hamcrest. I don't think we should. at least on these kind of simple assertions the standard Junit ones are perfectly readable to me. :|

PR #63, opened 2018-01-22
PP-3048 Update API documents

SandorArpa:
- update `READ.md` and `api_specification.md` to include the new endpoint


docs/api_specification.md
```diff
@@ -339,3 +339,39 @@ Content-Type: application/json
...
+
+### Request example
+
+```
+PATCH /v1/api/gateway-account/56
+Authorization: Bearer API_TOKEN
+Content-Type: application/json
+
+{
+    "op"    :     "replace",
```
daniele-occhipinti: Why we need to specify and operation? Isn't PATCH always about updating? For deleting we have the DELETE method.

SandorArpa: it could be `disable`, `enable`, etc

PR #76, opened 2018-03-07
PP-3441 Insecure direct object reference vulnerability

ckalista:
While creating new ` Prototype Links ` and ` Payment Links `in selfservice we were vulnerable to insecure direct object references as we failed to verify that products belong to gateway accounts for certain operations. To mitigate this problem, product API has been updated to explicitly force to provide the gateway account id in the path for all operations. 

* Substituted all variables in resource paths by their value to make it more
readable what the actual path for a resource is. This is consistent with the 
approach taken in other `Dropwizard` project such as Connector. 

* Added a method called `findByGatewayAccountIdAndExternalId` in `ProductDao` that restraints the lookup of products by their respective accounts. As part of this, a few missing tests have also been retrofitted. 

* Added a method called `findByGatewayAccountIdAndExternalId` in `ProductFinder` that restraints the lookup of products by their respective accounts.

* Added new endpoints mimicking the exact same functional behaviour of  existing endpoints to find/disable products but constraint by gateway accounts. Prevented insecure object reference for the following operations:
	* Find a product by external id *and gateway account id*, `/gateway-account/{gatewayAccountId}/products/{productExternalId}`
	* Disable a product by external id *and gateway account id*`/gateway-account/{gatewayAccountId}/products/{productExternalId}/disable`
	* Find all active products by *gateway account id*`/gateway-account/{gatewayAccountId}/products`
Note: The old endpoint have been kept provisionally for backward compatibility. A subsequent PR will remove them after final merge.

### DIFF
--- a/src/main/java/uk/gov/pay/products/model/Product.java
+++ b/src/main/java/uk/gov/pay/products/model/Product.java
@@ -127,4 +127,21 @@ public class Product {
     public String getReturnUrl() {
         return returnUrl;
     }
+
+    @Override
+    public String toString() {
+        return "Product{" +
+                "externalId='" + externalId + '\'' +
+                ", name='" + name + '\'' +
+                ", description='" + description + '\'' +
+                ", payApiToken='" + payApiToken + '\'' +
+                ", price=" + price +
+                ", status=" + status +
+                ", type=" + type +
+                ", gatewayAccountId=" + gatewayAccountId +
+                ", links=" + links +
+                ", returnUrl='" + returnUrl + '\'' +
+                ", serviceName='" + serviceName + '\'' +
+                '}';
+    }
 }
--- a/src/main/java/uk/gov/pay/products/model/Product.java
+++ b/src/main/java/uk/gov/pay/products/model/Product.java
@@ -134,7 +134,6 @@ public class Product {
                 "externalId='" + externalId + '\'' +
                 ", name='" + name + '\'' +
                 ", description='" + description + '\'' +
-                ", payApiToken='" + payApiToken + '\'' +
                 ", price=" + price +
                 ", status=" + status +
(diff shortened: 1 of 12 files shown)


src/main/java/uk/gov/pay/products/model/Product.java
```diff
@@ -127,4 +127,21 @@ public ProductType getType() {
...
         return returnUrl;
     }
+
+    @Override
+    public String toString() {
+        return "Product{" +
+                "externalId='" + externalId + '\'' +
+                ", name='" + name + '\'' +
+                ", description='" + description + '\'' +
+                ", payApiToken='" + payApiToken + '\'' +
```
SandorArpa: Do we want to expose the API token in the logs?

src/main/java/uk/gov/pay/products/resources/ProductResource.java
```diff
@@ -64,21 +60,58 @@ public Response findProduct(@PathParam("productExternalId") String productExtern
...
+
+    // TODO: TO BE REMOVED!
+    // Leaving this provisionally for backward compatibility
     @PATCH
-    @Path(DISABLE_PRODUCT_RESOURCE_PATH)
+    @Path("/products/{productExternalId}/disable")
     @Produces(APPLICATION_JSON)
     @Consumes(APPLICATION_JSON)
-    public Response disableProduct(@PathParam("productExternalId") String productExternalId) {
+    public Response deprecatedDisableProduct(@PathParam("productExternalId") String productExternalId) {
```
SandorArpa: Shouldn't we add @Deprecated annotation as well?

ckalista: Addressed

PR #78, opened 2018-03-11
PP-3441 Reinstated endpoints previously deprecated

ckalista:
The new endpoints constraining all product operations by gateway accounts introduced in c49ca1e0abaac1386e3f4b2471a277a4fcda1b53, also deprecated the original endpoints. As it became apparent that the original endpoints were also needed, this change reinstates them and adds missing tests.

# NOTE
To be merged only after:
- https://github.com/alphagov/pay-selfservice/pull/703
- https://github.com/alphagov/pay-products-ui/pull/67

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,10 +23,12 @@ The JSON naming convention follows Hypertext Application Language (HAL).
 | Path                          | Supported Methods | Description                        |
 | ----------------------------- | ----------------- | ---------------------------------- |
 |[```/v1/api/products```](docs/api_specification.md#post-v1apiproducts)        | POST    |  Creates a new product definition            |
-|[```/v1/api/products/{productId}```](docs/api_specification.md#get-v1apiproductsproductid)        | GET    |  Gets an existing product with the specified Id   |
-|[```/v1/api/products?gatewayAccountId={gatewayAccountId}```](docs/api_specification.md#get-v1apiproductsgatewayaccountid)        | GET    |  Gets lists of products that belongs to a gateway account   |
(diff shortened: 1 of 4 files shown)


daniele-occhipinti:
@ckalista are we planning to remove the "unsafe" endpoints, e.g. findProductByExternalId

ckalista:
@daniele-occhipinti No, we need to keep all the endpoints. The reason is we only have an gateway account id for the selfservice journeys. When it comes to paying user journeys we don't know anything about geteway accounts, all we have is an product external id or a payment external id in most situations hence why we need to keep the "unsafe" endpoints.

daniele-occhipinti:
@ckalista I see. Maybe it is a bit confusing for a user of the `products` API understand why they should use one over the other (passing `gatewayId` vs not passing it). We need to mention the security rationale at least in the docs? Btw, the README file needs to be updated for the new endpoints, right?

ckalista:
I have updated the README, thanks for reminding of that. Not sure there is anything to explain regarding the rationale of adding new endpoints though. Depending on what is needed people have the choice.

daniele-occhipinti:
That's cool.
What I meant is this. Say in one year time, we are not around, and somebody needs to use the products API to delete a product, from selfservice or any other microservice. They look at the docs and they see 2 endpoints which do exactly the same thing - one of them requires an extra param `gatewayId`. They will probably just use the simpler one, and so they will not use the security protection, because they don't see the point of using the more complex one, unless we clearly write in the docs why the more complex is available.
Am I making any sense? Maybe not?

ckalista:
Yes it makes sense @daniele-occhipinti. I have added a line to give people a clue as to why we have added these two new endpoints.

PR #80, opened 2018-03-12
PP-3308 Service look up by friendly url

SandorArpa:
- Add friendly url to ProductEntity and Product model
- Refactor endpoint to search by either friendly url or external id
- Refactor/add tests

### DIFF
--- a/src/main/java/uk/gov/pay/products/persistence/dao/ProductDao.java
+++ b/src/main/java/uk/gov/pay/products/persistence/dao/ProductDao.java
@@ -17,13 +17,13 @@ public class ProductDao extends JpaDao<ProductEntity> {
         super(entityManager, ProductEntity.class);
     }
 
-    public Optional<ProductEntity> findByExternalId(String externalId) {
+    public Optional<ProductEntity> findByFriendlyUrlOrExternalId(String friendlyUrlOrExternalId) {
         String query = "SELECT product FROM ProductEntity product " +
-                "WHERE product.externalId = :externalId";
+                "WHERE product.friendlyUrl = :friendlyUrlOrExternalId OR product.externalId = :friendlyUrlOrExternalId";
 
         return entityManager.get()
                 .createQuery(query, ProductEntity.class)
-                .setParameter("externalId", externalId)
+                .setParameter("friendlyUrlOrExternalId", friendlyUrlOrExternalId)
                 .getResultList().stream().findFirst();
     }
 
--- a/src/main/java/uk/gov/pay/products/persistence/entity/ProductEntity.java
+++ b/src/main/java/uk/gov/pay/products/persistence/entity/ProductEntity.java
@@ -48,6 +48,9 @@ public class ProductEntity extends AbstractEntity {
     @Column(name = "service_name")
     private String serviceName;
 
+    @Column(name = "friendly_url")
+    private String friendlyUrl;
+
     public ProductEntity() {
     }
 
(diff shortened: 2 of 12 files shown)


src/main/java/uk/gov/pay/products/persistence/entity/ProductEntity.java
```diff
@@ -48,6 +48,9 @@
     @Column(name = "service_name")
     private String serviceName;
 
+    @Column(name = "friendly_url")
+    private String friendlyUrl;
```
ckalista: Although in Product what we return should be the full friendly URL, what is held here in the entity is not the full friendly URL but merely a portion of it hence why we should probably settle for a different naming.

src/main/java/uk/gov/pay/products/persistence/dao/ProductDao.java
```diff
@@ -17,13 +17,13 @@ protected ProductDao(Provider<EntityManager> entityManager) {
         super(entityManager, ProductEntity.class);
     }
 
-    public Optional<ProductEntity> findByExternalId(String externalId) {
+    public Optional<ProductEntity> findByFriendlyUrlOrExternalId(String friendlyUrlOrExternalId) {
         String query = "SELECT product FROM ProductEntity product " +
-                "WHERE product.externalId = :externalId";
+                "WHERE product.friendlyUrl = :friendlyUrlOrExternalId OR product.externalId = :friendlyUrlOrExternalId";
```
ckalista: We should discuss this as I wonder what should be returned if we have more than one match. We should probably make sure this can never happen, nevertheless should take the best course of action should it happen.

src/main/java/uk/gov/pay/products/resources/ProductResource.java
```diff
@@ -54,7 +54,20 @@ public Response createProduct(JsonNode payload) {
     @Consumes(APPLICATION_JSON)
     public Response deprecatedFindProduct(@PathParam("productExternalId") String productExternalId) {
         logger.info("Find a product with externalId - [ {} ]", productExternalId);
```
ckalista: If you rebase you will find that this method is not deprecated anymore hence we probably want to update it again...

src/test/java/uk/gov/pay/products/service/ProductFinderTest.java
```diff
@@ -138,6 +138,32 @@ public void disableByGatewayAccountIdAndExternalId_shouldReturnEmpty_whenNotFoun
...
+        String externalId = "1";
+        String friendlyUrl = URLEncoder.encode("kent-council/pay-for-your-parking-permit", "UTF-8");
+        ProductEntity productEntity = new ProductEntity();
+        productEntity.setExternalId(externalId);
+        productEntity.setFriendlyUrl(friendlyUrl);
+        when(productDao.findByFriendlyUrlOrExternalId(friendlyUrl)).thenReturn(Optional.of(productEntity));
+
+        Optional<Product> productOptional = productFinder.findByFriendlyUrlOrExternalId(friendlyUrl);
+
+        assertThat(productOptional.isPresent(), is(true));
```
ckalista: If you rebase, you will find a new product matcher that you may want to use.

src/main/java/uk/gov/pay/products/persistence/dao/ProductDao.java
SandorArpa: It could never happen as both friendlyUrl and external id are unique

PR #97, opened 2018-04-09
PP-3556 Add DELETE endpoint to ProductResource

SandorArpa:
- add new endpoint to ProductResource
- add new method to ProductFinder (deleteByExternalId)
- add relevant tests


src/main/java/uk/gov/pay/products/service/ProductFinder.java
```diff
@@ -43,6 +44,16 @@ public ProductFinder(ProductDao productDao, LinksDecorator linksDecorator) {
                 .orElseGet(Optional::empty);
     }
 
+    @Transactional
+    public Optional<Boolean> deleteByExternalId(String externalId) {
```
ckalista: Why returning Optional<Boolean> here?

PR #99, opened 2018-04-10
PP-3126: Remove things needed for PAAS

whpearson:
- Removed manifest.yml, used for deployment
 - rds-combined-ca-bundle was used for trusting rds in PAAS
 - Removed line adding it in the Dockerfile


daniele-occhipinti:
Shall we remove this manifest too? https://github.com/alphagov/pay-products-ui/blob/master/manifest.yml

whpearson:
@daniele-occhipinti : https://github.com/alphagov/pay-products-ui/pull/79

PR #115, opened 2018-08-20
PP-4095 Ignore unknown properties when deserialising payment

alexbishop1:
When deserialising a payment response from public API into a `PaymentResponse` object, ignore any unknown properties that are encountered in the JSON.

This makes it possible to add new properties to the payment responses returned by public API without having to modify this project.

### DIFF
--- a/src/test/java/uk/gov/pay/products/stubs/publicapi/PublicApiStub.java
+++ b/src/test/java/uk/gov/pay/products/stubs/publicapi/PublicApiStub.java
@@ -100,6 +100,7 @@ public class PublicApiStub {
                                                 .add("href", "https://an.example.link/from/payment/platform")
                                                 .add("method", "GET")))
                 .add("card_brand", "Visa")
+                .add("Do not add this property to PaymentResponse", "To test deserialisation handles new/unknown properties")
                 .build();
     }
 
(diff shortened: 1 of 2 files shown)


src/test/java/uk/gov/pay/products/stubs/publicapi/PublicApiStub.java
```diff
@@ -100,6 +101,7 @@ public static JsonObject createPaymentResponsePayload(String paymentId, long amo
                                                 .add("href", "https://an.example.link/from/payment/platform")
                                                 .add("method", "GET")))
                 .add("card_brand", "Visa")
+                .add(RandomStringUtils.randomAscii(8), "To test deserialisation handles new/unknown properties")
```
rhowe-gds: This seems unnecessarily nondeterministic. Can't we just have a fixed string that we will never add to our API?

alexbishop1: I was worried someone might add the fixed string to `PaymentRequest`!

alexbishop1: But in general I’m in favour of deterministic tests, yes.

src/test/java/uk/gov/pay/products/stubs/publicapi/PublicApiStub.java
```diff
@@ -1,5 +1,6 @@
 package uk.gov.pay.products.stubs.publicapi;
 
+import org.apache.commons.lang3.RandomStringUtils;
```
rhowe-gds: Unused import

PR #124, opened 2018-11-05
PP-4297 fix response objects to ignore unknown fields

SandorArpa:
## WHAT
Fix response objects to ignore unknown fields. This is needed as we're extending publicapi responses and they're breaking products.

## HOW
add `@JsonIgnoreProperties(ignoreUnknown = true)` to relevant classes

### DIFF
--- a/src/main/java/uk/gov/pay/products/client/publicapi/model/Address.java
+++ b/src/main/java/uk/gov/pay/products/client/publicapi/model/Address.java
@@ -1,10 +1,12 @@
 package uk.gov.pay.products.client.publicapi.model;
 
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.PropertyNamingStrategy;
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/pay/products/client/publicapi/PaymentRequest.java
```diff
@@ -1,10 +1,12 @@
 package uk.gov.pay.products.client.publicapi;
 
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.PropertyNamingStrategy;
 import com.fasterxml.jackson.databind.annotation.JsonNaming;
 
 @JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy.class)
+@JsonIgnoreProperties(ignoreUnknown = true)
```
alexbishop1: We shouldn’t need this here: we never deserialise anything into a `PaymentRequest` object, so we don’t have to worry about unknown properties

PR #167, opened 2019-01-04
PP-4584: Manage HTTP proxies via env. vars

rhowe-gds:
Remove all code for configuring HTTP proxies and instead use environment
variables to set JVM system properties.

This makes the application respond to the following 3 environment variables:

http_proxy - passed as http.proxyHost
https_proxy - passed as https.proxyHost
java_http_non_proxy_hosts - passed as http.nonProxyHosts


docker-startup.sh
```diff
@@ -4,10 +4,14 @@ set -eu
 RUN_MIGRATION=${RUN_MIGRATION:-false}
 RUN_APP=${RUN_APP:-true}
 
-java -jar *-allinone.jar waitOnDependencies *.yaml && \
+[ -z "${http_proxy:-}" ] || JAVA_OPTS="${JAVA_OPTS:-} -Dhttp.proxyHost=${http_proxy}"
+[ -z "${https_proxy:-}" ] || JAVA_OPTS="${JAVA_OPTS:-} -Dhttps.proxyHost=${https_proxy}"
+[ -z "${java_http_non_proxy_hosts:-}" ] || JAVA_OPTS="${JAVA_OPTS:-} -Dhttp.nonProxyHosts=${java_http_non_proxy_hosts}"
+
```
danworth: Seems like we were setting `https.proxyPort` with the line that is being removed from `ProductsApplication.java` 
`System.setProperty("https.proxyPort", proxyConfiguration.getPort().toString());`
I don't see this being set here, is it no longer needed?

oswaldquek: yeah it's no longer needed. it will be part of the `http_proxy` string.

rhowe-gds: actually I lied. Dan is correct :(

PR #172, opened 2019-01-11
BAU: More IntelliJ-inspired tidyups

rhowe-gds:
* Use try with resources to manage the lifecycle of JDBC `Connection` objects
* Use IntelliJ's preferred `L` designator for `long` literals
* Remove an unused constructor in `DropwizardAppWithPostgresRule`
* Tighten a couple of access levels
* Turn an unused field into a method variable

### DIFF
--- a/src/test/java/uk/gov/pay/products/infra/DropwizardAppWithPostgresRule.java
+++ b/src/test/java/uk/gov/pay/products/infra/DropwizardAppWithPostgresRule.java
@@ -1,6 +1,5 @@
 package uk.gov.pay.products.infra;
 
-import com.google.inject.persist.jpa.JpaPersistModule;
 import io.dropwizard.db.DataSourceFactory;
 import io.dropwizard.testing.ConfigOverride;
 import io.dropwizard.testing.junit.DropwizardAppRule;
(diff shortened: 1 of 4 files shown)


src/test/java/uk/gov/pay/products/infra/DropwizardAppWithPostgresRule.java
```diff
@@ -102,12 +93,10 @@ public DatabaseTestHelper getDatabaseTestHelper() {
...
 
     private void registerShutdownHook() {
-        Runtime.getRuntime().addShutdownHook(new Thread(() -> {
-            postgres.stop();
-        }));
+        Runtime.getRuntime().addShutdownHook(new Thread(postgres::stop));
     }
 
-    private JpaPersistModule createJpaModule(final PostgresDockerRule postgres) {
+    private void createJpaModule(final PostgresDockerRule postgres) {
```
danworth: This looked a bit suspicious, it's used in a single place which does nothing with the response hence Intellij say's make it `void`. It doesn't look like the method has any lasting effect, it creates a `JpaPersistModule` but doesn't register it or anything. I commented out the single place its used and all the tests passed, so I think this method can be removed (to be safe I consulted Krishna and George who also think it looks unnecessary so probably safe..)

PR #175, opened 2019-01-15
BAU: Yet more IntelliJ-inspired cleanups

rhowe-gds:
* We can make use of `Objects`'s `equals` and `hashCode` methods to make our life easier
* One of our tests was using random numbers as input. This is not reliable.
* Remove some unused bits and bobs

### DIFF
--- a/src/main/java/uk/gov/pay/products/util/RandomIdGenerator.java
+++ b/src/main/java/uk/gov/pay/products/util/RandomIdGenerator.java
@@ -1,7 +1,5 @@
 package uk.gov.pay.products.util;
 
-import java.math.BigInteger;
-import java.security.SecureRandom;
 import java.util.Random;
 import java.util.UUID;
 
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/pay/products/util/RandomIdGenerator.java
```diff
@@ -18,21 +15,6 @@
...
-     * This works by choosing 130 bits from a cryptographically secure random bit generator,
-     * and encoding them in base-32.
-     * <p> 128 bits is considered to be cryptographically strong,
-     * but each digit in a base 32 number can encode 5 bits, so 128 is rounded up to the next multiple of 5.
-     * This encoding is compact and efficient, with 5 random bits per character. Compare this to a random UUID,
-     * which only has 3.4 bits per character in standard layout, and only 122 random bits in total </p>
-     *
-     * @return a random number in base32 (in string format)
-     */
-    public static String newId() {
```
alexbishop1: I assume this isn’t used anywhere

PR #185, opened 2019-01-23
PP-4660: Use LoggingFilter from pay-java-commons

rhowe-gds:
pay-java-commons has an equivalent LoggingFilter we can use to replace our own.


rhowe-gds:
7

PR #166, opened 2019-01-04
BAU: Upgrade mockserver from 5.5.0 to 5.5.1

rhowe-gds:

PR #206, opened 2019-02-27
bau: deprecate mockserver

oswaldquek:
This commit deprecates https://github.com/alphagov/pay-products/pull/166. When
upgrading mockserver, some integration tests were failing. However each test in
that integration test was passing when run in isolation, but some failed when
all the tests were run. Couldn't find a reason as to why that was happening, so
switching to a more reliable mocking library (wiremock).


rhowe-gds:
I reran the job in case it was a transient issue, but I'm not convinced it is
```
13:07:20 Tests in error: 
13:07:20   uk.gov.pay.products.client.publicapi.PublicApiRestClientTest: java.lang.RuntimeException: java.io.IOException: Failed to bind to /0.0.0.0:8080
```

PR #208, opened 2019-03-05
PP-4925 Remove service name from products

SandorArpa:
## WHAT YOU DID
- Remove service name from products. It will be served from adminusers.
- Remove `PATCH` service name endpoint.

This is part of the custom branding on payment links epic. We're going to query adminusers for the custom branding and use the service name from the same query. This ensures that service name is up to date.

## How to test

- How should it be reviewed? -> 
No reference to service name exists (not to be confused with service name path).
No references to `PATCH` service name exists.

### Documentation
Removed references to service name.
Removed references to `PATCH` service name.


src/main/java/uk/gov/pay/products/resources/GatewayAccountResource.java
```diff
@@ -1,54 +0,0 @@
...
-    private final GatewayAccountFactory gatewayAccountFactory;
-
-    @Inject
-    public GatewayAccountResource(GatewayAccountRequestValidator requestValidator, GatewayAccountFactory gatewayAccountFactory) {
-        this.requestValidator = requestValidator;
-        this.gatewayAccountFactory = gatewayAccountFactory;
-    }
-
-    @PATCH
-    @Path("/gateway-account/{gatewayAccountId}")
```
DanailMinchev: Just to confirm: we do not need this endpoint anymore because it was used for service name updates only and we do not want to have gateway account update functionality anymore?

SandorArpa: Yes, that's correct. We stored service name in the original design as `products` was supposed to be running in PaaS environment. Then we moved `products` to our AWS environment but never updated the code. Keeping service name in sync is a bit messy. When we decided that we're going to call `adminusers` for the service object so we can implement custom branding, we also took the opportunity to use the same service object as the source of truth. Hence there is no need to keep updating the service name (which was already out of sync, anyway).

PR #211, opened 2019-03-14
PP-4926 drop unused column service_name from products

SandorArpa:
## WHAT YOU DID
- drop unused column


src/main/resources/migrations/00031_drop_column_service_name_from_products.sql
```diff
@@ -0,0 +1,5 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:drop_column_service_name_on_products
+UPDATE products SET service_name = null;
```
stephencdaly: is set null before dropping necessary?

PR #218, opened 2019-03-22
PP-4864: Use the default Java truststore for TLS

rhowe-gds:
Prior to this, we ignored the default Java truststore and built our own using
the certificates from the directory specified by the CERTS_PATH environment
variable.

We can instead rely on the default Java truststore, although there are cases
where we need to add additional CA certificates:

* Amazon RDS has its own CA certificate. We add this to the default Java
  truststore when building our base container image.
* During our end to end tests, we supply custom certificates in order to run
  TLS in our docker-compose environment.

Handle these custom certificates in the existing way, using CERTS_PATH to point
to them and have the shell wrapper add them to the default Java truststore
prior to starting the application.

Since this is a rather slow operation, taking about 0.5s per certificate, have
some logic to avoid adding certificates to the truststore if they're already in
there, matching on their certificate fingerprint. We need to add openssl to the
container in order to extract that data.

By default, the PostgreSQL JDBC driver does not use the default Java truststore
to validate server certificates. Instead, it defaults to looking for
${user.home}/.postgresql/root.crt

Rather than supply this file specifically, we can use the provided
org.postgresql.ssl.DefaultJavaSSLFactory which will consult the default Java
truststore instead.


rhowe-gds:
PR builder currently failing with
`2019-03-28T10:34:53.745786761Z keytool error: java.io.FileNotFoundException: /etc/ssl/certs/java/cacerts (No such file or directory)`

PR #220, opened 2019-03-26
PP-4987 Add validations for language field in ProductRequest

stephencdaly:
- Add uk.gov.pay.model dependency to access SupportedLanguage
- Remove unused "service_name" validations
- Add validation for valid enum value and re-use for validating "type" is a valid ProductType

### Logging

- [ ] only emit log lines at ERROR level which require immediate attention from a support engineer. These will trigger a zendesk alert.

### Documentation

- [ ] Updated README.md for any of the following ?

* Introduced any new environment variables / removed existing environment variable
* Added new API / updated existing API definition

PR #221, opened 2019-03-26
PP-4987 Add language field to product

stephencdaly:
Add language field to ProductEntity and Product so that it is accepted by the create product endpoint, persisted to the database, and returned by the get product endpoints.

Default language to "en" when deserialising Product from JSON payload if the field is not present.

### Logging

### Documentation

### DIFF
--- a/src/main/java/uk/gov/pay/products/model/Product.java
+++ b/src/main/java/uk/gov/pay/products/model/Product.java
@@ -6,11 +6,15 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.PropertyNamingStrategy;
 import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import com.fasterxml.jackson.databind.ser.std.ToStringSerializer;
+import uk.gov.pay.commons.model.SupportedLanguage;
 import uk.gov.pay.products.util.ProductStatus;
 import uk.gov.pay.products.util.ProductType;
 
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/pay/products/model/Product.java
```diff
@@ -118,10 +137,11 @@ public static Product from(JsonNode jsonPayload) {
         Boolean referenceEnabled = (jsonPayload.get(FIELD_REFERENCE_ENABLED) != null) && jsonPayload.get(FIELD_REFERENCE_ENABLED).asBoolean();
         String referenceLabel = (jsonPayload.get(FIELD_REFERENCE_LABEL) != null) ? jsonPayload.get(FIELD_REFERENCE_LABEL).asText() : null;
         String referenceHint = (jsonPayload.get(FIELD_REFERENCE_HINT) != null) ? jsonPayload.get(FIELD_REFERENCE_HINT).asText() : null;
+        SupportedLanguage language = (jsonPayload.get(FIELD_LANGUAGE) != null) ? SupportedLanguage.fromIso639AlphaTwoCode(jsonPayload.get(FIELD_LANGUAGE).asText()) : SupportedLanguage.ENGLISH;
```
alexbishop1: I suppose an alternative might be something like this:

```java
SupportedLanguage language = Optional.ofNullable(jsonPayload.get(FIELD_LANGUAGE)).map(JsonNode::asText).map(SupportedLanguage::fromIso639AlphaTwoCode).orElse(SupportedLanguage.ENGLISH);
```

But consistency is good too

stephencdaly: I have made this change, it is nicer

stephencdaly:
> How does validation work here? If I send `"language": "fr"` (which isn’t supported), what happens?

Validation for the POST request was added in https://github.com/alphagov/pay-products/pull/220

PR #222, opened 2019-03-27
PP-5004 Send product language when creating a payment

stephencdaly:
Provide the language of the product when calling Public API to create the payment and expect it in the Public API response.

### Logging

### Documentation


src/main/java/uk/gov/pay/products/client/publicapi/PaymentRequest.java
```diff
@@ -55,13 +61,22 @@ public void setReturnUrl(String returnUrl) {
...
+    }
+
     @Override
     public String toString() {
         return "PaymentRequest{" +
                 "amount=" + amount +
                 ", reference='" + reference + '\'' +
                 ", description='" + description + '\'' +
                 ", returnUrl='" + returnUrl + '\'' +
+                ", language='" + language + '\'' +
```
alexbishop1: `language.toString()` will give output (`en` rather than `ENGLISH`)

PR #251, opened 2019-04-26
PP-4493 Expect Reference information in update product request

stephencdaly:
- Selfservice now sends the reference fields when making a request to update a product. Use this when updating the product

### Logging

### Documentation

### DIFF
--- a/src/test/java/uk/gov/pay/products/service/ProductCreatorTest.java
+++ b/src/test/java/uk/gov/pay/products/service/ProductCreatorTest.java
@@ -68,7 +68,7 @@ public class ProductCreatorTest {
         assertThat(product.getPayApiToken(), is(payApiToken));
         assertThat(product.getGatewayAccountId(), is(gatewayAccountId));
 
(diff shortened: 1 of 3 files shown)


src/test/java/uk/gov/pay/products/service/ProductCreatorTest.java
```diff
@@ -147,14 +152,60 @@ public void doUpdateByGatewayAccountId_shouldUpdateProduct() {
         verify(mockedProductEntity,  times(1)).setName(updatedName);
         verify(mockedProductEntity,  times(1)).setDescription(updatedDescription);
         verify(mockedProductEntity,  times(1)).setPrice(updatedPrice);
+        verify(mockedProductEntity, times(1)).setReferenceEnabled(true);
```
alexbishop1: I know you’re just following the conventions of the file but `times(1)` is unnecessary because it’s the default

stephencdaly: 👍

PR #271, opened 2019-05-15
PP-5191 No e2e test on PRs by default

SandorArpa:
As agreed with team, will make builds faster and reduce strain on CI

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -3,6 +3,10 @@
 pipeline {
   agent any
 
+  parameters {
+    booleanParam(defaultValue: false, description: '', name: 'runEndToEndTestsOnPR')
+  }
+
   options {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -14,6 +14,7 @@ pipeline {
   environment {
     DOCKER_HOST = "unix:///var/run/docker.sock"
     AWS_DEFAULT_REGION = "eu-west-1"
+    RUN_END_TO_END_ON_PR = "${params.runEndToEndTestsOnPR}"
```
stephencdaly: do we need to add the parameter in a `parameters { ... }` block?

and we need to add a conditional to check to decide whether to run the end-to-end stage depending on the value of the parameter

PR #318, opened 2019-08-28
Remove parameter to tagDeployment() in Jenkinsfile

rhowe-gds:
Since https://github.com/alphagov/pay-jenkins-library/pull/117, tagDeployment()
needs no parameters.


rhowe-gds:
We had to revert the `tagDeployment()` change :(

PR #414, opened 2020-02-18
PP-6149 add product metadata validation

SandorArpa:
## WHAT YOU DID
- keys have to be unique (case insensitive) compared to existing keys
- only one key-value pair is allowed
- keys are limited to 20 characters
- values are limited to 50 characters

- add tests

### DIFF
--- a/src/main/java/uk/gov/pay/products/validations/ProductsMetadataRequestValidator.java
+++ b/src/main/java/uk/gov/pay/products/validations/ProductsMetadataRequestValidator.java
@@ -1,0 +1,56 @@
+package uk.gov.pay.products.validations;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.google.inject.Inject;
+import uk.gov.pay.products.model.ProductMetadata;
+import uk.gov.pay.products.util.Errors;
+
+import java.util.List;
+import java.util.Optional;
+
+import static java.lang.String.format;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/products/validations/ProductsMetadataRequestValidator.java
```diff
@@ -0,0 +1,56 @@
...
+        if (existingMetadataList
+                .stream()
+                .map(ProductMetadata::getKey)
+                .map(String::toLowerCase)
+                .anyMatch(isEqual(key.toLowerCase()))) {
+            return Optional.of(Errors.from(format("Key [ %s ] already exists, duplicate keys not allowed", key)));
+        }
+
+        if (key.length() > MAX_KEY_FIELD_LENGTH) {
+            return Optional.of(Errors.from(format("Maximum key field length is [ %s ]", MAX_KEY_FIELD_LENGTH)));
```
sfount: Other validations with fixed values have static fields for the messages (`MAX_NUMBER_OF_METADATA_ALLOWED_ERROR_MSG`), for consistency the key and value field lengths could do this too

SandorArpa: I'll change it in the next PR

PR #415, opened 2020-02-18
PP-6165 List Metadata For Payment Link

rorymalcolm:
- When a payment link is returned as JSON, the metadata that relates to that payment link instance should be contained within the response object under the `metadata` tag.

### DIFF
--- a/src/main/java/uk/gov/pay/products/persistence/entity/ProductMetadataEntity.java
+++ b/src/main/java/uk/gov/pay/products/persistence/entity/ProductMetadataEntity.java
@@ -2,7 +2,6 @@ package uk.gov.pay.products.persistence.entity;
 
 import uk.gov.pay.products.model.ProductMetadata;
 
-import javax.persistence.CascadeType;
 import javax.persistence.Column;
 import javax.persistence.Entity;
 import javax.persistence.JoinColumn;
@@ -13,9 +12,9 @@ import javax.persistence.Table;
 @Table(name = "products_metadata")
 public class ProductMetadataEntity extends AbstractEntity {
 
-    @ManyToOne(cascade = CascadeType.MERGE)
-    @JoinColumn(name = "product_id", updatable = false)
-    private ProductEntity productId;
+    @ManyToOne()
+    @JoinColumn(name = "product_id", referencedColumnName = "id")
+    private ProductEntity productEntity;
 
     @Column(name = "metadata_key")
     private String metadataKey;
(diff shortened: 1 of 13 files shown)


src/test/java/uk/gov/pay/products/resources/ProductResourceIT.java
```diff
@@ -918,4 +921,50 @@ public void shouldReturn400_whenSavingProduct_withReferenceEnabledAndNoReference
...
+        String productsUrl = "https://products.url/v1/api/products/";
+        String productsUIPayUrl = "https://products-ui.url/pay/";
+        response
+                .body("_links", hasSize(2))
+                .body("_links[0].href", matchesPattern(productsUrl + externalId))
+                .body("_links[0].method", is(HttpMethod.GET))
+                .body("_links[0].rel", is("self"))
+                .body("_links[1].href", matchesPattern(productsUIPayUrl + externalId))
+                .body("_links[1].method", is(HttpMethod.GET))
+                .body("_links[1].rel", is("pay"));
```
kbottla: Assertion for metadata ? as test specifically says `shouldReturnMetadata`

src/main/java/uk/gov/pay/products/persistence/entity/ProductMetadataEntity.java
```diff
@@ -13,9 +13,9 @@
 @Table(name = "products_metadata")
 public class ProductMetadataEntity extends AbstractEntity {
 
-    @ManyToOne(cascade = CascadeType.MERGE)
-    @JoinColumn(name = "product_id", updatable = false)
-    private ProductEntity productId;
+    @ManyToOne(fetch = FetchType.EAGER)
```
kbottla: any reason this is EAGER ? We seem to always get productEntity first to start processing (querying) metadata. Not sure if JPA issues additional queries to get productEntity for each metadata row

src/test/java/uk/gov/pay/products/resources/ProductResourceIT.java
rorymalcolm: exists on line 957, can remove this at is comes from another test

rorymalcolm: `                .body(METADATA + "[0].key",is("value"));`

src/main/java/uk/gov/pay/products/persistence/entity/ProductMetadataEntity.java
rorymalcolm: no reason, I can remove it if you think that's a better idea

PR #417, opened 2020-02-19
PP-6167 add products metadata updater endpoint

SandorArpa:
## WHAT YOU DID
- add new patch endpoint
- refactor validator
- add new dao method
- add new factory worker
- add relevant tests

### DIFF
--- a/src/main/java/uk/gov/pay/products/persistence/dao/ProductMetadataDao.java
+++ b/src/main/java/uk/gov/pay/products/persistence/dao/ProductMetadataDao.java
@@ -6,6 +6,7 @@ import uk.gov.pay.products.persistence.entity.ProductMetadataEntity;
 
 import javax.persistence.EntityManager;
 import java.util.List;
+import java.util.Optional;
 
 public class ProductMetadataDao extends JpaDao<ProductMetadataEntity> {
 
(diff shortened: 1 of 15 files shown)


src/main/java/uk/gov/pay/products/persistence/dao/ProductMetadataDao.java
```diff
@@ -33,4 +34,15 @@ public ProductMetadataDao(Provider<EntityManager> entityManager) {
                 .setParameter("productExternalId", productExternalId)
                 .getResultList();
     }
+
+    public Optional<ProductMetadataEntity> findByProductsExternalIdAndKey(String productExternalId, String key) {
+        String query = "SELECT metadata FROM ProductMetadataEntity metadata, ProductEntity product " +
+                "WHERE metadata.productEntity = product AND product.externalId = :productExternalId " +
```
sfount: I'm not 100% familiar with how this would work in JPA land but do we need to specify the join condition in the WHERE if there is a foreign key?

src/main/java/uk/gov/pay/products/persistence/dao/ProductMetadataDao.java
```diff
@@ -33,4 +34,15 @@ public ProductMetadataDao(Provider<EntityManager> entityManager) {
...
+                "AND lower(metadata.metadataKey) = :key";
```
sfount: The index on key is a union with `product_id` `(product_id, metadata_key)` -- I think this should be able to use it but just making sure

src/main/java/uk/gov/pay/products/resources/ProductMetadataResource.java
```diff
@@ -44,11 +42,33 @@ public Response createMetadata(@PathParam("productExternalId") String productExt
...
+                            .map(errors -> Response.status(Response.Status.BAD_REQUEST).entity(errors).build())
+                            .orElseGet(() -> {
+                                metadataFactory.metadataCreator().createProductMetadata(payload, productExternalId);
+                                return Response.status(Response.Status.CREATED).build();
+                            }));
+    }
+
+    @PATCH
+    @Path("/products/{productExternalId}/metadata")
+    @Produces(APPLICATION_JSON)
```
sfount: I think this is the same for all metadata routes - could it be applied to the resource instead of individual requests?

src/main/java/uk/gov/pay/products/service/ProductsMetadataCreator.java
```diff
@@ -32,7 +30,7 @@ public ProductMetadata createProductMetadata(ProductMetadata metadata, String pr
                     metadataEntity.setMetadataValue(metadata.getValue());
                     metadataDao.persist(metadataEntity);
                     return metadataEntity.toMetadata();
-                }).orElseThrow(() -> new ProductNotFoundException(format("Product with %s id not found", productExternalId), productExternalId));
+                }).orElseThrow(() -> new ProductNotFoundException(productExternalId));
```
sfount: Nice :+1:

src/main/java/uk/gov/pay/products/resources/ProductMetadataResource.java
SandorArpa: 👍

PR #436, opened 2020-03-09
PP-6157 Add Delete Endpoint For Metadata

rorymalcolm:
- Adds a DELETE endpoint for metadata which allows for metadata on a
product to be removed via an API.

- Adds tests to ensure happy and failure paths both work as expected.

### DIFF
--- a/src/main/java/uk/gov/pay/products/resources/ProductMetadataResource.java
+++ b/src/main/java/uk/gov/pay/products/resources/ProductMetadataResource.java
@@ -9,6 +9,7 @@ import uk.gov.pay.products.service.ProductsMetadataFactory;
 import uk.gov.pay.products.validations.ProductsMetadataRequestValidator;
 
 import javax.ws.rs.Consumes;
+import javax.ws.rs.DELETE;
 import javax.ws.rs.PATCH;
 import javax.ws.rs.POST;
 import javax.ws.rs.Path;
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/products/resources/ProductMetadataResource.java
```diff
@@ -69,4 +69,14 @@ public Response updateMetadata(@PathParam("productExternalId") String productExt
...
     }
+
+    @DELETE
+    @Path("/{metadataKey}")
+    public Response deleteMetadata(@PathParam("productExternalId") String productExternalId, @PathParam("metadataKey") String metadataKey) {
+        logger.info("Delete product metadata for id - [ {} ] with key id [ {} ]", productExternalId, metadataKey);
+
+        return metadataFactory.metadataDeleter().deleteMetadata(productExternalId, metadataKey) ?
+                Response.status(Response.Status.OK).build() :
+                Response.status(Response.Status.NOT_FOUND).build();
```
sfount: Does this `NOT_FOUND` builder actually get called or does this pass because the method throws `MetadataNotFoundException` ? 

If the latter I think this could be removed as I think there's no case where it's not found and has returned false

rorymalcolm: done!

PR #449, opened 2020-03-31
PP-6325 Indexes for product usage reports

sfount:
* index product type to allow the report to ignore demo and prototype
links
* index payment date_created to allow aggregate `MAX` in report

### DIFF
--- a/src/main/resources/migrations/00035_index_columns_for_usage_report.sql
+++ b/src/main/resources/migrations/00035_index_columns_for_usage_report.sql
@@ -1,0 +1,9 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:index_type_products runInTransaction:false
+CREATE INDEX CONCURRENTLY products_type_idx ON products(type);
+-- rollback drop index concurrently products_type_idx
+
+--changeset uk.gov.pay:index_product_id_date_created_payments runInTransaction:false
+CREATE INDEX CONCURRENTLY payments_product_id_date_created_idx ON payments(product_id, date_created);
+-- rollback drop index concurrently payments_product_id_date_created_idx


src/main/resources/migrations/00035_index_columns_for_usage_report.sql
```diff
@@ -0,0 +1,9 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:index_type_products
+CREATE INDEX products_type_idx ON products(type);
+-- rollback drop index products_type_idx
+
+--changeset uk.gov.pay:index_date_created_payments
+CREATE INDEX payments_date_created_idx ON payments(date_created);
```
kbottla: If query uses max(date_created) group by product_id, we also need to add product_id to the index.

kbottla: CREATE INDEX CONCURRENTLY ?

src/main/resources/migrations/00035_index_columns_for_usage_report.sql
```diff
@@ -0,0 +1,9 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:index_type_products
+CREATE INDEX products_type_idx ON products(type);
```
kbottla: CREATE INDEX CONCURRENTLY ?

src/main/resources/migrations/00035_index_columns_for_usage_report.sql
```diff
@@ -0,0 +1,9 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:index_type_products runInTransaction:false
+CREATE INDEX CONCURRENTLY products_type_idx ON products(type);
+-- rollback drop index concurrently products_type_idx
+
+--changeset uk.gov.pay:index_date_created_payments runInTransaction:false
+CREATE INDEX CONCURRENTLY payments_date_created_idx ON payments(product_id, date_created);
```
kbottla: Do we still want to use same name for the index ?

src/main/resources/migrations/00035_index_columns_for_usage_report.sql
```diff
@@ -0,0 +1,9 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:index_type_products runInTransaction:false
+CREATE INDEX CONCURRENTLY products_type_idx ON products(type);
+-- rollback drop index concurrently products_type_idx
+
+--changeset uk.gov.pay:index_date_created_payments runInTransaction:false
```
kbottla: name of the change still says only created_date but not major

PR #531, opened 2020-09-04
PP-6913 Change secret paths and configure hooks

rorymalcolm:
- Configure secret paths and initialise gds pre-commit


rjbaker (approved):
LGTM 🚀

PR #546, opened 2020-09-14
PP-7067 Increase maximum metadata value length from 50 to 100

alexbishop1:
Update `ProductsMetadataRequestValidator` (and its unit tests) to rely on constants in ExternalMetadata (which is imported from Pay Java commons) for things like the maximum number of characters allowed for metadata values associated with a product.

This has no functional effect other than increasing the character limit for a metadata value from 50 to 100. The necessary database migration to enable this was commit 32ae783331377b719910ab91df1b521a0f12d59f.

Public API and connector have already been updated to allow 100 characters rather than 50.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": null,
     "lines": null
   },
-  "generated_at": "2020-09-08T08:03:19Z",
+  "generated_at": "2020-09-14T12:31:04Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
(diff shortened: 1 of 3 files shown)


oswaldquek (approved):
Would prefer if `ExternalMetadata.[MAX_KEY_LENGTH|MAX_VALUE_LENGTH]` were imported statically but it's no biggie.

alexbishop1:
I did think about statically importing the constants from `ExternalMetadata` but `ProductsMetadataRequestValidator` already has its own constants, so I thought it might be unclear where the `ExternalMetadata` ones were coming from.

PR #565, opened 2020-10-23
PP-7293: Add a method to update Product's API key

adityapahuja:
Add a new method to enable ProductFinder to update Product's API key in a database using the given product external id. This will allow the key to be rotated immediately if it is compromised.

### DIFF
--- a/src/main/java/uk/gov/pay/products/service/ProductFinder.java
+++ b/src/main/java/uk/gov/pay/products/service/ProductFinder.java
@@ -46,6 +46,17 @@ public class ProductFinder {
     }
 
     @Transactional
+    public Optional<Product> updatePayApiTokenByExternalId(String externalId, String payApiToken) {
+        return productDao.findByExternalId(externalId)
+                .map(productEntity -> {
+                    productEntity.setPayApiToken(payApiToken);
+                    productDao.merge(productEntity);
+                    return Optional.of(productEntity.toProduct());
+                })
+                .orElseGet(Optional::empty);
+    }
+
+    @Transactional
     public Boolean deleteByExternalId(String externalId) {
         return productDao.findByExternalId(externalId)
                 .map(productEntity -> {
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/products/service/ProductFinder.java
```diff
@@ -45,6 +45,17 @@ public ProductFinder(ProductDao productDao, LinksDecorator linksDecorator) {
                 .orElseGet(Optional::empty);
     }
 
+    @Transactional
+    public Optional<Product> updatePayApiTokenByExternalId(String externalId, String payApiToken) {
+        return productDao.findByExternalId(externalId)
+                .map(productEntity -> {
+                    productEntity.setPayApiToken(payApiToken);
+                    productDao.merge(productEntity);
```
alexbishop1, line 53: I think the `@Transactional` means you don’t need to explicitly merge but it won’t do any harm.

alexbishop1 (approved):

src/main/java/uk/gov/pay/products/service/ProductFinder.java
adityapahuja: I thought`@Transactional` provides commit and rollback support. It would be nice to test this function against a real product DAO rather than the mocked product DAO.  Using the mocked product DAO in the unit test doesn't give me confidence that it will work without using merge() function.

adityapahuja: In other words, the unit test doesn't actually test this function using a real mini database.

adityapahuja: I have tested the `updatePayApiTokenByExternalId` function against a real product DAO. It works fine. When I remove `productDao.merge(productEntity);` from the function, it fails the test. The merge() function is required.

adityapahuja:
I have made further changes to remove the unused variable (productEntity) from `updatePayApiTokenByExternalId_shouldReturnEmpty_whenNotFound()` test and rename the variable from `disabledProduct` to `product` in `ProductFinderTest`.

alexbishop1 (approved):

PR #570, opened 2020-10-28
PP-7293: Add an endpoint to regenerate an api key for a product

adityapahuja:
## What you did
This pull request has three separate commits.
 
1. The first commit adds a new endpoint `v1/api/products/{productId}/regenerate-api-key` to the products resource. It is used to generate a new API token and replaces an old API key with the new key for a product with a specified external id. This endpoint is needed for rotating API key for a product because the key can be compromised. It adds two configs items (`emailAddress` and `publicAuthUrl`). Email address is required for identifying who makes a request for generating a new API key. Public Auth URL address is needed to allow Products application to send a request to Public Authentication application for regenerating a new API token. It adds integration tests to cover the following scenarios.

    * API key for payment link successfully replaced
    * API key for prototype link successfully replaced
    * API key for demo payment successfully replaced
    * New API key cannot be generated
    * Product doesn’t exist
    * API Key is null
    * API Key is empty

2. The second commit adds a unit test case to cover a scenario where a new API key cannot be updated. It is difficult to make Products application to fail to update API key in the database after receiving new API token in the integration test. Unit test allows us to do this.

3. The third commit adds information about two new environment variables and the new endpoint (`/v1/api/products/{productExternalId}/regenerate-api-key`) to README and API specification documentation.

## How to test

- How should it be reviewed? 
It is best to review each commit and run all tests to make sure that there are no failures.

- If manual testing is needed, give suggested testing steps.
Integration and unit tests cover the scenarios mentioned above. It would be nice to test these changes in a development environment.

### Logging

- [x] only emit log lines at ERROR level which require immediate attention from a support engineer. These will trigger a zendesk alert.

### Documentation

Yes, the following two environment variables are added. Please see the first commit. 
    * EMAIL_ADDRESS_FOR_ROTATING_API_KEYS
    * PUBLICAUTH_URL

Yes, please see the third commit.

### DIFF
--- a/src/main/java/uk/gov/pay/products/client/publicauth/TokenResponse.java
+++ b/src/main/java/uk/gov/pay/products/client/publicauth/TokenResponse.java
@@ -1,0 +1,35 @@
+package uk.gov.pay.products.client.publicauth;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.PropertyNamingStrategy;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+
+@JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy.class)
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class TokenResponse {
+    private String token;
+    
+    public TokenResponse() {
+    }
+
+    public TokenResponse(
+            @JsonProperty("token") String token) {
+        this.token = token;
+    }
+
+    public String getToken() {
+        return token;
+    }
+
+    public void setToken(String token) {
+        this.token = token;
+    }
+
+    @Override
+    public String toString() {
+        return "TokenResponse{" +
+                "token='" + token + '\'' +
+                '}';
+    }
+}
--- a/src/main/java/uk/gov/pay/products/model/CreateTokenRequest.java
+++ b/src/main/java/uk/gov/pay/products/model/CreateTokenRequest.java
@@ -1,0 +1,59 @@
+package uk.gov.pay.products.model;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import uk.gov.pay.commons.model.TokenPaymentType;
+
+import javax.validation.constraints.NotNull;
+
+import static java.util.UUID.randomUUID;
+import static uk.gov.pay.commons.model.TokenPaymentType.CARD;
+import static uk.gov.pay.products.model.TokenSource.API;
+
+public class CreateTokenRequest {
+
+    @NotNull private final String accountId;
+    @NotNull private final String description;
+    @NotNull private final String createdBy;
+    private final TokenPaymentType tokenPaymentType;
+    private final TokenSource tokenSource;
+    private final TokenLink tokenLink = TokenLink.of(randomUUID().toString());
+
+    @JsonCreator
+    public CreateTokenRequest(@JsonProperty("account_id") String accountId,
+                              @JsonProperty("description") String description,
+                              @JsonProperty("created_by") String createdBy,
+                              @JsonProperty("token_type") TokenPaymentType tokenPaymentType,
+                              @JsonProperty("type") TokenSource tokenSource) {
+        this.accountId = accountId;
+        this.description = description;
+        this.createdBy = createdBy;
+        this.tokenPaymentType = tokenPaymentType == null ? CARD : tokenPaymentType;
+        this.tokenSource = tokenSource == null ? API : tokenSource;
+    }
+
+    public String getAccountId() {
+        return accountId;
+    }
+
+    public String getDescription() {
+        return description;
+    }
+
+    public String getCreatedBy() {
+        return createdBy;
+    }
+
+    public TokenPaymentType getTokenPaymentType() {
+        return tokenPaymentType;
+    }
+
+    public TokenSource getTokenSource() {
+        return tokenSource;
+    }
+
+    public TokenLink getTokenLink() {
+        return tokenLink;
+    }
+}
--- a/src/main/java/uk/gov/pay/products/model/TokenSource.java
+++ b/src/main/java/uk/gov/pay/products/model/TokenSource.java
@@ -1,0 +1,18 @@
+package uk.gov.pay.products.model;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public enum TokenSource {
+    API, PRODUCTS;
+    private static final Logger LOGGER = LoggerFactory.getLogger(TokenSource.class);
+
+    public static TokenSource fromString(final String source) {
+        try {
+            return TokenSource.valueOf(source.toUpperCase());
+        } catch (Exception e) {
+            LOGGER.error("Unknown token source: " + source);
+            return API;
+        }
+    }
+}
--- a/src/main/java/uk/gov/pay/products/resources/ProductResource.java
+++ b/src/main/java/uk/gov/pay/products/resources/ProductResource.java
@@ -5,17 +5,26 @@ import com.google.inject.Inject;
 import io.dropwizard.jersey.PATCH;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import uk.gov.pay.products.client.publicauth.TokenResponse;
+import uk.gov.pay.products.config.ProductsConfiguration;
+import uk.gov.pay.products.model.CreateTokenRequest;
 import uk.gov.pay.products.model.Product;
 import uk.gov.pay.products.model.ProductUsageStat;
 import uk.gov.pay.products.service.ProductFactory;
 import uk.gov.pay.products.validations.ProductRequestValidator;
+
+import static uk.gov.pay.commons.model.TokenPaymentType.CARD;
 import static uk.gov.pay.logging.LoggingKeys.GATEWAY_ACCOUNT_ID;
 import static net.logstash.logback.argument.StructuredArguments.kv;
 
 import javax.ws.rs.*;
+import javax.ws.rs.client.Client;
+import javax.ws.rs.client.Entity;
+import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.Response.Status;
 import java.util.List;
+import java.util.Optional;
 
 import static java.lang.String.format;
 import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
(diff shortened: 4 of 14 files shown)


src/main/java/uk/gov/pay/products/resources/ProductResource.java
```diff
@@ -88,6 +105,38 @@ public Response findProductByGatewayAccountIdAndExternalId(@PathParam("gatewayAc
                         Response.status(NOT_FOUND).build());
     }
 
+    @POST
+    @Path("/products/{productExternalId}/regenerate-api-key")
+    @Produces(APPLICATION_JSON)
+    @Consumes(APPLICATION_JSON)
+    public Response regenerateProductApiKey(@PathParam("productExternalId") String productExternalId) {
+        return productFactory.productFinder().findByExternalId(productExternalId).map((product) -> {
```
alexbishop1, line 113: I think most of the logic here should be in a service rather than directly in the resource. The resource should really just call the service, which would do the actual work of regenerating the token, with the resource just producing the appropriate response based on whether or not it was successful.

A separate service would also be easier to unit test, which might mean we need fewer integration tests.

src/main/java/uk/gov/pay/products/resources/ProductResource.java
```diff
@@ -88,6 +105,38 @@ public Response findProductByGatewayAccountIdAndExternalId(@PathParam("gatewayAc
...
+            CreateTokenRequest createTokenRequest = new CreateTokenRequest(
+                    String.valueOf(product.getGatewayAccountId()),
+                    getDescription(product),
+                    emailAddress,
+                    CARD,
+                    PRODUCTS);
+
+            Response response = client.target(publicAuthUrl)
```
alexbishop1, line 121: I think `publicAuthUrl` is going to need the path adding to it somewhere since it’s just the hostname now (unless that happens somewhere else and I’ve missed it).

src/main/java/uk/gov/pay/products/resources/ProductResource.java
```diff
@@ -88,6 +105,38 @@ public Response findProductByGatewayAccountIdAndExternalId(@PathParam("gatewayAc
...
+                    .request()
+                    .post(Entity.entity(createTokenRequest, MediaType.APPLICATION_JSON));
+
+            TokenResponse tokenResponse = response.readEntity(TokenResponse.class);
+            String token;
+            if (tokenResponse != null && tokenResponse.getToken() != null && !tokenResponse.getToken().isEmpty()) {
```
alexbishop1, line 127: `Optional.ofNullable(tokenResponse).map(TokenResponse::getToken).filter(not(String::isEmpty)).map(…)` might make this is a little more readable.

But I question if we need the `isEmpty()` check? Do we really think public auth is likely to return an empty string?

src/main/java/uk/gov/pay/products/resources/ProductResource.java
```diff
@@ -183,4 +232,17 @@ public Response findProductsAndStats(@QueryParam("gatewayAccountId") Integer gat
...
+    private String getDescription(Product product) {
+        switch (product.getType()) {
+            case ADHOC:
+                return String.format("Token for \"%s\" payment link", product.getName());
+            case PROTOTYPE:
+                return String.format("Token for Prototype: %s", product.getName());
+            case DEMO:
+                return "Token for Demo Payment";
+            default:
+                return null;
```
alexbishop1, line 245: We should throw an exception if we reach here: something has gone terribly wrong.

src/test/java/uk/gov/pay/products/resources/ProductResourceIT.java
```diff
@@ -1035,4 +1045,371 @@ public void shouldReturnUsageStats_whenProductAndPaymensExist() {
                 .body("size", is(1))
                 .body("[0].product.external_id", is(product.getExternalId()));
     }
+
+    @Test
+    public void shouldReturn200_whenApiKeyForAPaymentLinkIsUpdated() {
```
alexbishop1, line 1050: Not sure we need a variant of this test for each of the different product types: we could probably do with just one and rely on unit tests to make sure we construct the correct request for each of the three types.

src/test/java/uk/gov/pay/products/resources/ProductResourceIT.java
```diff
@@ -1035,4 +1045,371 @@ public void shouldReturnUsageStats_whenProductAndPaymensExist() {
...
+                .get(format("/v1/api/products/%s", externalId))
+                .then()
+                .statusCode(200);
+
+        payApiToken = response.extract().path(PAY_API_TOKEN);
+        assertThat(payApiToken, is("default api key"));
+    }
+
+    @Test
+    public void shouldReturn500_whenAnApiKeyIsEmptyBeforeUpdatingApiKeyForAPaymentLink() {
```
alexbishop1, line 1274: Is this scenario realistic?

src/test/java/uk/gov/pay/products/resources/ProductResourceTest.java
```diff
@@ -0,0 +1,95 @@
...
+    @Before
+    public void setUp() {
+        when(productsConfiguration.getEmailAddress()).thenReturn("pay-products@gov.uk");
+        when(productsConfiguration.getPublicAuthUrl()).thenReturn(PUBLICAUTH_URL);
+
+        productResource = new ProductResource(productRequestValidator, productFactory, client, productsConfiguration);
+    }
+    
+    @Test
+    public void shouldReturn500_whenFailingToUpdateApiKeyForAPaymentLink() {
```
alexbishop1, line 59: Feels a bit odd that we have a unit test just for this one thing but everything else is tested by integration tests.

alexbishop1 (changes requested):
Basic approach looks good and as far as I can tell closely matches the acceptance criteria.

I think we could move some of the logic out of the resource and into a service so we‘re not so reliant on integration tests to test this (right now we’ve added more new integration tests than unit tests!).

There’s also some places where I feel it tries too hard to be resilient to errors but it would probably be better to just blow up (since we’re going to be the only ones calling this endpoint and will be ready to deal with the fallout).

This PR is also a little on the large side. Obviously it’s a bit late now but maybe we could try to split changes like this up into more incremental steps in the future.

src/main/java/uk/gov/pay/products/model/CreateTokenRequest.java
```diff
@@ -0,0 +1,59 @@
...
+
+import uk.gov.pay.commons.model.TokenPaymentType;
+
+import javax.validation.constraints.NotNull;
+
+import static java.util.UUID.randomUUID;
+import static uk.gov.pay.commons.model.TokenPaymentType.CARD;
+import static uk.gov.pay.products.model.TokenSource.API;
+
+public class CreateTokenRequest {
```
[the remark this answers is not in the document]
adityapahuja: This model class is taken from PublicAuth application. Is it fine to rename this class to `CreateApiTokenRequest`?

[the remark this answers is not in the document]
alexbishop1: Yeah, if it’s going to live within the products microservice, it makes sense for it to be named from the perspective of how products sees it.

src/main/java/uk/gov/pay/products/resources/ProductResource.java
alexbishop1: I don’t think we have to worry too much about intercepted responses between microservices given they’re all internal HTTPS calls. Even if products did store an empty string, it would mean the API calls it makes to public API would fail (because the API key wouldn’t be a working one).

adityapahuja: Thanks. I will drop `isEmpty()`.

src/main/java/uk/gov/pay/products/model/CreateTokenRequest.java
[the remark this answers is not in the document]
adityapahuja: Thanks. I will rename it.

src/main/java/uk/gov/pay/products/client/publicauth/TokenResponse.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    public void setToken(String token) {
+        this.token = token;
+    }
+
+    @Override
+    public String toString() {
+        return "TokenResponse{" +
+                "token='" + token + '\'' +
```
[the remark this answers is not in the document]
adityapahuja: I was using toString() for testing purposes. I will drop it.

src/main/java/uk/gov/pay/products/model/CreateTokenRequest.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @NotNull private final String accountId;
+    @NotNull private final String description;
+    @NotNull private final String createdBy;
```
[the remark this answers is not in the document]
adityapahuja: This model class is taken from Public Auth application. `@NotNull` signifies that these fields (account id, description and createdBy) are mandatory.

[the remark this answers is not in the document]
alexbishop1: I see. I think preserving that in the products copy is a little confusing if we’re not sharing literally the same file.

src/test/java/uk/gov/pay/products/resources/ProductResourceIT.java
adityapahuja: I have dropped it.

src/test/java/uk/gov/pay/products/resources/ProductResourceTest.java
adityapahuja: This is a special case where a product is deleted from the database just before Products application replaces an old API token with a new API token for the product. It is difficult to do this in integration test. That's why I am able to use mock in unit test to create this kind of scenario.

src/main/java/uk/gov/pay/products/resources/ProductResource.java
adityapahuja: It is in the config file.

src/main/java/uk/gov/pay/products/model/CreateTokenRequest.java
```diff
@@ -0,0 +1,59 @@
...
+    private final TokenPaymentType tokenPaymentType;
+    private final TokenSource tokenSource;
+    private final TokenLink tokenLink = TokenLink.of(randomUUID().toString());
```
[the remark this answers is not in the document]
adityapahuja: Good question. I am not sure why Public Auth application wants `TokenLink` in CreateTokenRequest model class.

src/main/java/uk/gov/pay/products/model/TokenSource.java
```diff
@@ -0,0 +1,18 @@
...
+public enum TokenSource {
+    API, PRODUCTS;
+    private static final Logger LOGGER = LoggerFactory.getLogger(TokenSource.class);
+
+    public static TokenSource fromString(final String source) {
+        try {
+            return TokenSource.valueOf(source.toUpperCase());
+        } catch (Exception e) {
+            LOGGER.error("Unknown token source: " + source);
+            return API;
```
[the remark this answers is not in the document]
adityapahuja: Good idea. I will make this change.

adityapahuja:
This pull request has been superseded by a new pull request (https://github.com/alphagov/pay-products/pull/573).

src/main/java/uk/gov/pay/products/model/TokenSource.java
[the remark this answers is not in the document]
adityapahuja: fromString() method is not used. I can drop it.

PR #572, opened 2020-10-29
PP-7205 Use commons version of RestClientLoggingFilter

stephencdaly:
This has been moved to pay-java-commons to reduce duplication. This also
fixes a bug where we were using the wrong key to get the request id from
the MDC, so we weren't logging it.


kbottla (approved):
LGTM!

PR #573, opened 2020-10-30
PP-7293: Add a service for handling API token for a product

adityapahuja:
## WHAT YOU DID
This change

- adds a new service called Product API Token Manager which is responsible for getting a new API token from Public Auth application and replacing an old API token with a new API token for a specified product.
- introduces two configuration items (`emailAddressForReplacingApiTokens` and `publicAuthUrl`). `emailAddressForReplacingApiTokens` is required for identifying who makes a request for generating a new API token. `publicAuthUrl` is needed by Products application to send a request to Public Auth application for generating a new API token.
- add unit tests to ensure that the service can receive a new API token from the Public Auth application and replace an old API token with a new API token in a database for a given product.

At this moment, Products application is not using this service. There will be a separate pull request for adding a new endpoint which will use this service. This pull request should not be merged until another pull request for adding two environment variables (EMAIL_ADDRESS_FOR_REPLACING_API_TOKENS and PUBLICAUTH_URL) is merged. 

### Logging

- [X ] only emit log lines at ERROR level which require immediate attention from a support engineer. These will trigger a zendesk alert.

### Documentation

There will be a pull request for adding a new endpoint for replacing API token for a product. It will include changes to README and API specifications.
 
Yes, the following two environment variables are added. 
    * EMAIL_ADDRESS_FOR_REPLACING_API_KEYS
    * PUBLICAUTH_URL

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -18,7 +18,9 @@
         <eclipselink.version>2.7.7</eclipselink.version>
         <guice.version>4.2.3</guice.version>
         <hamcrest.version>2.2</hamcrest.version>
+        <junit5.version>5.7.0</junit5.version>
         <mainClass>uk.gov.pay.products.ProductsApplication</mainClass>
+        <mockito.version>3.5.15</mockito.version>
         <powermock.version>2.0.7</powermock.version>
         <wiremock.version>2.27.2</wiremock.version>
         <jackson.version>2.11.2</jackson.version>
@@ -204,7 +206,31 @@
         <dependency>
             <groupId>org.mockito</groupId>
             <artifactId>mockito-core</artifactId>
-            <version>3.5.15</version>
+            <version>${mockito.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-junit-jupiter</artifactId>
+            <version>${mockito.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
+            <version>${junit5.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-engine</artifactId>
+            <version>${junit5.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.vintage</groupId>
+            <artifactId>junit-vintage-engine</artifactId>
+            <version>${junit5.version}</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/src/main/java/uk/gov/pay/products/client/publicauth/model/CreateApiTokenRequest.java
+++ b/src/main/java/uk/gov/pay/products/client/publicauth/model/CreateApiTokenRequest.java
@@ -1,0 +1,83 @@
+package uk.gov.pay.products.client.publicauth.model;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import uk.gov.pay.commons.model.TokenPaymentType;
+
+import java.util.Objects;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class CreateApiTokenRequest {
+    @JsonProperty("account_id")
+    private String accountId;
+
+    @JsonProperty("description")
+    private String description;
+
+    @JsonProperty("created_by")
+    private String createdBy;
+
+    @JsonProperty("token_type")
+    private TokenPaymentType tokenPaymentType;
+
+    @JsonProperty("type")
+    private TokenSource tokenSource;
+
+    public CreateApiTokenRequest() {}
+
+    public CreateApiTokenRequest(String accountId,
+                                 String description,
+                                 String createdBy,
+                                 TokenPaymentType tokenPaymentType,
+                                 TokenSource tokenSource) {
+        this.accountId = accountId;
+        this.description = description;
+        this.createdBy = createdBy;
+        this.tokenPaymentType = tokenPaymentType;
+        this.tokenSource = tokenSource;
+    }
+
+    public String getAccountId() {
+        return accountId;
+    }
+
+    public String getDescription() {
+        return description;
+    }
+
+    public String getCreatedBy() {
+        return createdBy;
+    }
+
+    public TokenPaymentType getTokenPaymentType() {
+        return tokenPaymentType;
+    }
+
+    public TokenSource getTokenSource() {
+        return tokenSource;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+
+        if (!(o instanceof CreateApiTokenRequest)) {
+            return false;
+        }
+
+        CreateApiTokenRequest that = (CreateApiTokenRequest) o;
+
+        return accountId.equals(that.accountId) &&
+                description.equals(that.description) &&
+                createdBy.equals(that.createdBy) &&
+                tokenPaymentType == that.tokenPaymentType &&
+                tokenSource == that.tokenSource;
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(accountId, description, createdBy, tokenPaymentType, tokenSource);
+    }
+}
--- a/src/main/java/uk/gov/pay/products/client/publicauth/model/NewApiTokenFromPublicAuthResponse.java
+++ b/src/main/java/uk/gov/pay/products/client/publicauth/model/NewApiTokenFromPublicAuthResponse.java
@@ -1,0 +1,23 @@
+package uk.gov.pay.products.client.publicauth.model;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.PropertyNamingStrategy;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+
+@JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy.class)
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class NewApiTokenFromPublicAuthResponse {
+    @JsonProperty
+    private String token;
+
+    public NewApiTokenFromPublicAuthResponse() {}
+
+    public NewApiTokenFromPublicAuthResponse(String token) {
+        this.token = token;
+    }
+
+    public String getToken() {
+        return token;
+    }
+}
--- a/src/main/java/uk/gov/pay/products/service/ProductApiTokenManager.java
+++ b/src/main/java/uk/gov/pay/products/service/ProductApiTokenManager.java
@@ -1,0 +1,87 @@
+package uk.gov.pay.products.service;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import uk.gov.pay.products.client.publicauth.model.CreateApiTokenRequest;
+import uk.gov.pay.products.client.publicauth.model.NewApiTokenFromPublicAuthResponse;
+import uk.gov.pay.products.config.ProductsConfiguration;
+import uk.gov.pay.products.exception.FailToGetNewApiTokenException;
+import uk.gov.pay.products.exception.FailToReplaceApiTokenException;
+import uk.gov.pay.products.model.Product;
+
+import javax.inject.Inject;
+import javax.ws.rs.client.Client;
+import javax.ws.rs.client.Entity;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import java.util.Optional;
+
+import static uk.gov.pay.commons.model.TokenPaymentType.CARD;
+import static uk.gov.pay.products.client.publicauth.model.TokenSource.PRODUCTS;
+
+public class ProductApiTokenManager {
+    private static final Logger LOGGER = LoggerFactory.getLogger(ProductApiTokenManager.class);
+    public static final String NEW_API_TOKEN_PATH = "/v1/frontend/auth";
+
+    private final Client client;
+    private final String publicAuthUrl;
+    private final String emailAddressForReplacingApiTokens;
+    private final ProductFactory productFactory;
+
+    @Inject
+    public ProductApiTokenManager(Client client, ProductsConfiguration configuration, ProductFactory productFactory) {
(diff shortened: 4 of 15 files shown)


src/main/java/uk/gov/pay/products/client/publicauth/model/CreateApiTokenRequest.java
```diff
@@ -0,0 +1,61 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import uk.gov.pay.commons.model.TokenPaymentType;
+
+import javax.validation.constraints.NotNull;
+
+import static java.util.UUID.randomUUID;
+import static uk.gov.pay.commons.model.TokenPaymentType.CARD;
+import static uk.gov.pay.products.client.publicauth.model.TokenSource.PRODUCTS;
+
+public class CreateApiTokenRequest {
```
alexbishop1, line 13: I understand from comments on https://github.com/alphagov/pay-products/pull/570 that this was copied directly from public auth. I think we’re fine to let this diverge and represent the request as products sees it rather than as public auth sees it. For example, public auth’s idea of what’s mandatory and what to default to don’t really affect how products uses it.

(If we were using this request in many places then I could see the value of keeping the class consistent everywhere — and having it live in Pay Java commons — but right now I don’t think we’re using it enough to make the overhead worth it.)

src/main/java/uk/gov/pay/products/client/publicauth/model/NewApiTokenFromPublicAuthResponse.java
```diff
@@ -0,0 +1,26 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+
+@JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy.class)
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class NewApiTokenFromPublicAuthResponse {
+    private String token;
+
+    public NewApiTokenFromPublicAuthResponse() {}
+
+    public NewApiTokenFromPublicAuthResponse(@JsonProperty("token") String token) {
```
alexbishop1, line 15: Probably a bit more conventional to have the annotation on the field.

src/main/java/uk/gov/pay/products/exception/FailToRotateApiTokenException.java
```diff
@@ -0,0 +1,8 @@
+package uk.gov.pay.products.exception;
+
+public class FailToRotateApiTokenException extends RuntimeException {
```
alexbishop1, line 3: I’m being a bit picky here but maybe `FailToReplaceApiTokenException`? To me, “rotate” implies the old token is also revoked.

src/main/java/uk/gov/pay/products/service/ProductApiTokenManager.java
```diff
@@ -0,0 +1,82 @@
...
+import javax.ws.rs.client.Entity;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import java.util.Optional;
+
+import static uk.gov.pay.commons.model.TokenPaymentType.CARD;
+import static uk.gov.pay.products.client.publicauth.model.TokenSource.PRODUCTS;
+
+public class ProductApiTokenManager {
+    private final Logger logger = LoggerFactory.getLogger(getClass());
```
alexbishop1, line 24: This can be `private static final Logger LOGGER`.

src/main/java/uk/gov/pay/products/service/ProductApiTokenManager.java
```diff
@@ -0,0 +1,82 @@
...
+
+    private Client client;
```
alexbishop1, line 26: I assume this can be `final` too.

src/main/java/uk/gov/pay/products/service/ProductApiTokenManager.java
```diff
@@ -0,0 +1,82 @@
...
+                        String.format("Failed to get a new API token for product %s of type %s", 
+                                product.getExternalId(),
+                                product.getType())));
+    }
+
+    public void rotateApiTokenForAProduct(Product product, String newApiToken) {
+        productFactory.productFinder().updatePayApiTokenByExternalId(product.getExternalId(), newApiToken).ifPresentOrElse(modifiedProduct -> 
+                        logger.info(String.format("Regenerated API token for product %s of type %s", modifiedProduct.getExternalId(), modifiedProduct.getType())),
+                () -> { throw new FailToRotateApiTokenException(
+                        String.format("Failed to rotate API token for product %s of type %s", product.getExternalId(), product.getType())); });
```
alexbishop1, line 58: Don’t need the curly braces here.

src/main/java/uk/gov/pay/products/service/ProductApiTokenManager.java
```diff
@@ -0,0 +1,82 @@
...
+    private String getDescription(Product product) {
+        switch (product.getType()) {
+            case ADHOC:
+                return String.format("Token for \"%s\" payment link", product.getName());
+            case PROTOTYPE:
+                return String.format("Token for Prototype: %s", product.getName());
+            case DEMO:
+                return "Token for Demo Payment";
+            default:
+                throw new InvalidProductTypeException(String.format("Invalid product type [%s]", product.getType().toString()));
```
alexbishop1, line 79: I think having a specific `InvalidProductTypeException` for something that should never happen is probably overkill. It could just be an `IllegalArgumentException`.

src/main/resources/config/config.yaml
```diff
@@ -74,8 +74,11 @@ jpa:
 jerseyClientConfiguration:
   disabledSecureConnection: "false"
 
+emailAddressForRotatingApiTokens: ${EMAIL_ADDRESS_FOR_ROTATING_API_TOKENS}
+
 baseUrl: ${BASE_URL}
 publicApiUrl: ${PUBLICAPI_URL}
+publicAuthUrl: ${PUBLICAUTH_URL}/v1/frontend/auth
```
alexbishop1, line 81: Maybe `publicAuthNewTokenUrl` or something? Just because everywhere else `microserviceNameUrl` is a URL without a path.

src/test/java/uk/gov/pay/products/service/ProductApiTokenManagerTest.java
```diff
@@ -0,0 +1,132 @@
...
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertThrows;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+@RunWith(MockitoJUnitRunner.class)
```
alexbishop1, line 39: I’m not going to insist you change this since it’s written now but we can use JUnit 5 for new unit test classes.

src/test/java/uk/gov/pay/products/service/ProductApiTokenManagerTest.java
```diff
@@ -0,0 +1,132 @@
...
+
+    private Appender<ILoggingEvent> mockAppender;
+
+    @Captor
+    private ArgumentCaptor<LoggingEvent> loggingEventArgumentCaptor;
+
+    @Before
+    public void setUp() {
+        Logger root = (Logger) LoggerFactory.getLogger(ProductApiTokenManager.class);
+        mockAppender = mock(Appender.class);
```
alexbishop1, line 65: Is there something about the fact this has to be added to `root` that stops us using the `@Mock` annotation for this?

src/test/java/uk/gov/pay/products/service/ProductApiTokenManagerTest.java
```diff
@@ -0,0 +1,132 @@
...
+                is(String.format("Failed to rotate API token for product %s of type %s", demoPayment.getExternalId(), demoPayment.getType())));
+    }
+
+    private void setUpPublicAuthToReturnResponse(String expectedNewApiToken) {
+        WebTarget webTarget = mock(WebTarget.class);
+        when(client.target(PUBLICAUTH_URL)).thenReturn(webTarget);
+        Invocation.Builder builder = mock(Invocation.Builder.class);
+        when(webTarget.request()).thenReturn(builder);
+        Response publicAuthResponse = mock(Response.class);
+        when(builder.post(any(Entity.class))).thenReturn(publicAuthResponse);
```
alexbishop1, line 128: Can we make this a bit more rigorous? It would be good to check we’re sending the right payload (including the description) for each of the three types of products. We could have three variants of `shouldReturnANewApiTokenWhenPublicAuthGeneratesTheToken` to facilitate this.

alexbishop1 (changes requested):
This is looking nicer than doing everything in the resource!

src/main/java/uk/gov/pay/products/client/publicauth/model/CreateApiTokenRequest.java
adityapahuja: I have dropped TokenLink. It is not a mandatory field.

src/main/java/uk/gov/pay/products/exception/FailToRotateApiTokenException.java
adityapahuja: Done.

src/test/java/uk/gov/pay/products/service/ProductApiTokenManagerTest.java
adityapahuja: Yes, I have made this a bit more rigorous.

src/test/java/uk/gov/pay/products/service/ProductApiTokenManagerTest.java
adityapahuja: I have added a separate commit for updating ProductApiTokenManagerTest to use JUnit5.

src/test/java/uk/gov/pay/products/service/ProductApiTokenManagerTest.java
```diff
@@ -0,0 +1,182 @@
...
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+import static uk.gov.pay.products.service.ProductApiTokenManager.NEW_API_TOKEN_PATH;
+
+@ExtendWith(MockitoExtension.class)
+public class ProductApiTokenManagerTest {
+    private static final String PUBLICAUTH_URL = "https://publicauth.url";
+    private static final String EMAIL_ADDRESS_FOR_REPLACING_API_TOKENS = "pay-products@gov.uk";
+    private static final String NEW_API_TOKEN = "New API Token";
```
alexbishop1, line 48: Rather using the same new token, maybe we could have different one for each test? e.g. “New API token for payment link”, “New API token for demo payment” etc. Just to make what’s going on a little clearer.

alexbishop1 (changes requested):
I think this is almost there. Just a few small-ish comments.

pom.xml
```diff
@@ -18,7 +18,9 @@
         <eclipselink.version>2.7.7</eclipselink.version>
         <guice.version>4.2.3</guice.version>
         <hamcrest.version>2.2</hamcrest.version>
+        <junit5_version>5.7.0</junit5_version>
```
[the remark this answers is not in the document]
adityapahuja: I didn't notice a dot. This has been done.

src/main/java/uk/gov/pay/products/client/publicauth/model/CreateApiTokenRequest.java
```diff
@@ -0,0 +1,77 @@
...
+    public CreateApiTokenRequest(@JsonProperty("account_id") String accountId,
+                                 @JsonProperty("description") String description,
+                                 @JsonProperty("created_by") String createdBy,
+                                 @JsonProperty("token_type") TokenPaymentType tokenPaymentType,
+                                 @JsonProperty("type") TokenSource tokenSource) {
+        this.accountId = accountId;
+        this.description = description;
+        this.createdBy = createdBy;
+        this.tokenPaymentType = tokenPaymentType == null ? CARD : tokenPaymentType;
+        this.tokenSource = tokenSource == null ? PRODUCTS : tokenSource;
```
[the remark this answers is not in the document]
adityapahuja: This has been done.

src/main/java/uk/gov/pay/products/service/ProductApiTokenManager.java
```diff
@@ -0,0 +1,83 @@
...
+    public String getNewApiTokenFromPublicAuth(Product product) {
+        CreateApiTokenRequest tokenRequest = createApiTokenRequest(product);
+
+        Response response = client.target(publicAuthUrl + NEW_API_TOKEN_PATH)
+                .request()
+                .post(Entity.entity(tokenRequest, MediaType.APPLICATION_JSON));
+
+        return Optional.ofNullable(response.readEntity(NewApiTokenFromPublicAuthResponse.class))
+                .map(NewApiTokenFromPublicAuthResponse::getToken)
+                .orElseThrow(() -> new MissingApiTokenException(
```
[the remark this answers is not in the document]
adityapahuja: This has been done.

src/test/java/uk/gov/pay/products/service/ProductApiTokenManagerTest.java
adityapahuja: This has been done.

src/main/java/uk/gov/pay/products/service/ProductApiTokenManager.java
```diff
@@ -0,0 +1,83 @@
...
+                        String.format("Failed to get a new API token for product %s of type %s", 
+                                product.getExternalId(),
+                                product.getType())));
+    }
+
+    public void replaceApiTokenForAProduct(Product product, String newApiToken) {
+        productFactory.productFinder().updatePayApiTokenByExternalId(product.getExternalId(), newApiToken).ifPresentOrElse(modifiedProduct -> 
+                        LOGGER.info(String.format("Regenerated API token for product %s of type %s", modifiedProduct.getExternalId(), modifiedProduct.getType())),
+                () -> { throw new FailToReplaceApiTokenException(
+                        String.format("Failed to replace API token for product %s of type %s", product.getExternalId(), product.getType())); });
```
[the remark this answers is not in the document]
adityapahuja: This has been done.

pom.xml
```diff
@@ -18,7 +18,9 @@
...
         <mainClass>uk.gov.pay.products.ProductsApplication</mainClass>
+        <mockito_version>3.5.15</mockito_version>
```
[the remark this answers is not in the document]
adityapahuja: This has been done.

alexbishop1 (dismissed):
Looks good!

alexbishop1 (approved):

PR #574, opened 2020-10-30
PP-7293: Add an endpoint for replacing an old API token with a new API token for a product

adityapahuja:
## WHAT YOU DID
This pull request has two commits.

1. The first commit adds a new endpoint `v1/api/products/{productId}/regenerate-api-token` to the products resource. It is used to generate a new API token and replaces an old API token with the new token for a product with a specified external id. This endpoint is needed because the API token can be compromised. It also adds integration tests to ensure the the endpoint can return 200 for successfully replacing API token, 500 for failing to replacing it and 404 for the product which does not exist.
2. The second commit adds information about two new environment variables (EMAIL_ADDRESS_FOR_REPLACING_API_TOKENS and PUBLICAUTH_URL) and the new endpoint (/v1/api/products/{productExternalId}/regenerate-api-token) to README and API specification documentation.

### Logging

- [x] only emit log lines at ERROR level which require immediate attention from a support engineer. These will trigger a zendesk alert.

### Documentation

Yes
Yes

### DIFF
--- a/src/main/java/uk/gov/pay/products/resources/ProductResource.java
+++ b/src/main/java/uk/gov/pay/products/resources/ProductResource.java
@@ -7,22 +7,32 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.products.model.Product;
 import uk.gov.pay.products.model.ProductUsageStat;
+import uk.gov.pay.products.service.ProductApiTokenManager;
 import uk.gov.pay.products.service.ProductFactory;
 import uk.gov.pay.products.validations.ProductRequestValidator;
-import static uk.gov.pay.logging.LoggingKeys.GATEWAY_ACCOUNT_ID;
-import static net.logstash.logback.argument.StructuredArguments.kv;
 
-import javax.ws.rs.*;
+import javax.ws.rs.Consumes;
+import javax.ws.rs.DELETE;
+import javax.ws.rs.GET;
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/pay/products/resources/ProductResource.java
```diff
@@ -183,4 +197,15 @@ public Response findProductsAndStats(@QueryParam("gatewayAccountId") Integer gat
         List<ProductUsageStat> usageStats = productFactory.productFinder().findProductsAndUsage(gatewayAccountId);
         return Response.status(OK).entity(usageStats).build();
     }
+
+    @POST
+    @Path("/products/{productExternalId}/regenerate-api-token")
+    @Produces(APPLICATION_JSON)
+    @Consumes(APPLICATION_JSON)
+    public Response regenerateProductApiToken(@PathParam("productExternalId") String productExternalId) {
+        return productFactory.productFinder().findByExternalId(productExternalId).map((product) -> {
```
alexbishop1, line 206: Don’t really need the brackets around `(product)`.

alexbishop1 (approved):

src/main/java/uk/gov/pay/products/resources/ProductResource.java
adityapahuja: Will remove the brackets from a single parameter next time.

PR #596, opened 2020-12-11
PP-PP-7358 add metadata validator for product create request

SandorArpa:
## WHAT YOU DID
- add a metadata validator for product create request. this is needed as the requirements slightly differ to when adding metadata for an existing product
- add relevant tests


kbottla (approved):
LGTM!

PR #599, opened 2020-12-15
PP-7507 Process metadata when patching product

kbottla:
## WHAT YOU DID
- Adds functionality to validate and update metadata when PATCHing (editing) product.

with @sandorarpa


SandorArpa (approved):
LOKTM
I’m confident that this change does not require further review and meets the secure coding requirements of GOV.UK Pay

PR #625, opened 2021-02-16
BAU Don't log errors when 4xx exceptions are thrown

stephencdaly:
For exceptions that are mapped to a 4xx response don't log an error so they aren't reported in Sentry


SandorArpa (approved):
LGTM

PR #639, opened 2021-03-29
PP-7938: Use pay-java-commons from maven central

oswaldquek:


alexbishop1 (changes requested):
`src/main/resources/META-INF/persistence.xml` will need updating too.

oswaldquek:
> `src/main/resources/META-INF/persistence.xml` will need updating too.

Ooh thanks.

alexbishop1 (approved):

PR #670, opened 2021-05-26
PP-7921 Rollback Products Migration

rorymalcolm:
- Rollback a DB migration for products deployments to test the migration script


katstevens (approved):
Drop it like it's 🔥

PR #672, opened 2021-06-01
PP-8166 Remove card details response deserialiser

sfount:
Card brand and card details (including card brand) are never used by
products or products-ui.

The duplicated definition of card brands causes a significant bug when
deserialising unexpected card brand values.

Remove the Jackson POJO mappers that specify how to deserialise the
public API payment response card details message.

This fixed enum has no way of staying up to date with card connectors supported
types and should be reintroduced as a shared enum if it's actually required.


stephencdaly (approved):
🗑️

PR #697, opened 2021-07-13
PP-8210 - Add integration test for publicapi returning 403 for credentials not configured

nimalank7:
Description:
- Publicapi returns 403 in the event that the account is not fully configured. The PaymentCreator then throws a PublicApiResponseErrorException which is caught and mapped to a PaymentCreationException. This then returns 500 with 'Downstream system error'. Nothing further needs to be done to handle the 403 except adding an integration test to handle this case.


src/test/java/uk/gov/pay/products/resources/PaymentResourceIT.java
```diff
@@ -264,6 +264,40 @@ public void createAPayment_shouldFail_whenProductIsNotFound() {
                 .body("errors[0]", is(format("Product with product id %s not found.", unknownProductId)));
     }
 
+    @Test
+    public void createAPayment_shouldReturn404_whenPublicApiReturnsAccountNotLinked() {
```
kbottla, line 268:
```suggestion
    public void createAPayment_shouldReturn500_whenPublicApiReturnsAccountNotLinked() {
```

nimalank7: Done.

kbottla (approved):

PR #713, opened 2021-08-02
PP-8415 Detect and log PAN in reference field

mrlumbu:
Log when we see suspected PAN in reference field. This will enable us to remove `reference` from logs whilst still catching the case when users do this. Products does not log the reference field at all, but once we have this in place can remove the logging from publicapi and connector.

I have chosen to implement this in a separate class to make it easily testable. I have put it in the validator package which is a bit loose, but seems best place for it (happy to discuss this). 

For the actual criteria for "looking like a PAN" I have copied and improved on what we do in frontend [here](https://github.com/alphagov/pay-frontend/blob/437a67cb8c50a9368bb767d9c6b2ca5ce4dee784/app/utils/charge-validation-fields.js#L145) and [here](https://github.com/alphagov/pay-frontend/blob/437a67cb8c50a9368bb767d9c6b2ca5ce4dee784/app/utils/charge-validation-fields.js#L145) - length between 12 and 19 inclusive, Luhn valid.

### DIFF
--- a/src/main/java/uk/gov/pay/products/service/PaymentCreator.java
+++ b/src/main/java/uk/gov/pay/products/service/PaymentCreator.java
@@ -22,6 +22,7 @@ import uk.gov.pay.products.service.transaction.TransactionFlow;
 import uk.gov.pay.products.service.transaction.TransactionalOperation;
 import uk.gov.pay.products.util.PaymentStatus;
 import uk.gov.pay.products.util.ProductType;
+import uk.gov.pay.products.validations.PanDetector;
 import uk.gov.service.payments.commons.model.Source;
 
 import javax.inject.Inject;
@@ -70,16 +71,16 @@ public class PaymentCreator {
         return linksDecorator.decorate(paymentEntity.toPayment());
     }
 
-    private TransactionalOperation<TransactionContext, PaymentEntity> beforePaymentCreation(String productExternalId, String reference) {
+    private TransactionalOperation<TransactionContext, PaymentEntity> beforePaymentCreation(String productExternalId, String userDefinedReference) {
         return context -> {
             logger.info("Creating a new payment for product external id {}", productExternalId);
             ProductEntity productEntity = productDao.findByExternalId(productExternalId)
                     .orElseThrow(() -> new PaymentCreatorNotFoundException(productExternalId));
             if (productEntity.getReferenceEnabled()) {
-                if (isEmpty(reference)) {
+                if (isEmpty(userDefinedReference)) {
                     throw new BadPaymentRequestException("User defined reference is enabled but missing");
                 }
-                return mergePaymentEntityWithoutReferenceCheck(setupPaymentEntity(productEntity, reference));
+                return mergePaymentEntityWithoutReferenceCheck(setupPaymentEntity(productEntity, userDefinedReference));
             }
             
             PaymentEntity paymentEntity = setupPaymentEntity(productEntity, randomUserFriendlyReference());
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/products/service/PaymentCreator.java
```diff
@@ -70,16 +71,19 @@ public Payment doCreate(String productExternalId, Long priceOverride, String ref
...
             ProductEntity productEntity = productDao.findByExternalId(productExternalId)
                     .orElseThrow(() -> new PaymentCreatorNotFoundException(productExternalId));
             if (productEntity.getReferenceEnabled()) {
-                if (isEmpty(reference)) {
+                if (isEmpty(userDefinedReference)) {
                     throw new BadPaymentRequestException("User defined reference is enabled but missing");
                 }
-                return mergePaymentEntityWithoutReferenceCheck(setupPaymentEntity(productEntity, reference));
+                if (PANDetector.isSuspectedPAN(userDefinedReference)) {
+                    logger.warn("Suspected PAN entered by user in reference field");
```
stephencdaly, line 84: Would it be useful to do this after we've created the payment so we can log a payment external id?

src/test/java/uk/gov/pay/products/validations/PANDetectorTest.java
```diff
@@ -0,0 +1,38 @@
...
+        assertFalse(PANDetector.isSuspectedPAN("4242424242"));
+    }
+
+    @Test
+    public void shouldNotDetectPotentialPAN_whenNumberIsLongerThan19Digits() {
+        assertFalse(PANDetector.isSuspectedPAN("42424242424242424242"));
+    }
+    
+    @Test
+    public void shouldNotDetectPotentialPAN_whenNumberDoesNotPasLuhnAlgorithm() {
```
alexbishop1, line 35: Missing an `s` in `Pass`.

alexbishop1 (changes requested):
Looks good!

src/main/java/uk/gov/pay/products/validations/PANDetector.java
```diff
@@ -0,0 +1,19 @@
+package uk.gov.pay.products.validations;
+
+import org.apache.commons.validator.routines.checkdigit.LuhnCheckDigit;
+
+public class PANDetector {
+    private static final Integer MIN_PAN_LENGTH = 11;
+    private static final Integer MAX_PAN_LENGTH = 19;
```
alexbishop1, line 7: These can be primitive `int`s. It looks like we use a minimum of 12 not 11 in frontend: https://github.com/alphagov/pay-frontend/blob/437a67cb8c50a9368bb767d9c6b2ca5ce4dee784/app/utils/charge-validation-fields.js#L161

src/main/java/uk/gov/pay/products/service/PaymentCreator.java
mrlumbu: Good point. I'd have to do it after the payment has been created, and also will lose context as to whether this was a user created reference, but that's fine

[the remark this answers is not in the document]
mrlumbu: yep, agree, and similar to @stephencdaly's comment

src/main/java/uk/gov/pay/products/service/PaymentCreator.java
```diff
@@ -141,6 +142,11 @@ private PaymentEntity mergePaymentEntityWithReferenceNumberCheck(PaymentEntity p
                         kv(PAYMENT_EXTERNAL_ID, paymentEntity.getGovukPaymentId()),
                         kv("product_external_id", paymentEntity.getProductEntity().getExternalId())
                 );
+                
+                if (PanDetector.isSuspectedPan(paymentEntity.getReferenceNumber())) {
+                    logger.warn("Suspected PAN entered by user in reference field for payment id {}", 
+                            paymentResponse.getPaymentId());
```
stephencdaly, line 148: could this use structured logging?
`kv(PAYMENT_EXTERNAL_ID, externalId)`

mrlumbu: yep. can you tell I've been away from this for a while...

stephencdaly (approved):

alexbishop1 (approved):

PR #740, opened 2021-08-27
PP-8551 Allow PATCH require_captcha

stephencdaly:
Add a /v2/api/ endpoint that uses the JSON Patch specification to update a product.

This is required as the current /v1/ endpoint for updating a product expects all the fields that can be updated to be provided, and does not have the ability to update a single field. Adding the new parameter to this endpoint would risk properties of a product being overwritten if all consumers of this endpoint aren't kept in sync.

The new /v2/ endpoint is expected to be used for all updates in the future, but for now it will only support updating the `require_captcha` property.


src/main/java/uk/gov/pay/products/resources/ProductResource.java
```diff
@@ -156,8 +159,26 @@ public Response updateProduct(@PathParam("gatewayAccountId") Integer gatewayAcco
...
     }
 
+    /**
+     * Update product using JSON patch. 
+     * 
+     * TODO: Updates from /v1/api/gateway-account/{gatewayAccountId}/products/{productExternalId}
+     * should be moved over to this endpoint and that API deprecated. This was not done at the time due to time constraints.
+     */
+    @PATCH
+    @Path("/v2/api/products/{productExternalId}")
```
kbottla, line 169: Any reason we can't use `/v1/api/products/{productExternalId}` since this is unique ?

kbottla (commented):
Looks good overall. Added a comment about versioning api.

kbottla (approved):
LGTM!

PR #769, opened 2021-11-02
PP-8780: log gateway account ID and reference length for suspected PAN in reference

gidsg:
## WHAT YOU DID
Implemented enhanced logging to support additional filtering for this alert (option 3 as described in https://payments-platform.atlassian.net/browse/PP-8780)

## How to test

- How should it be reviewed? 
- What feedback do you need? 
- If manual testing is needed, give suggested testing steps.

## Code review checklist
Tests should pass
Code changes should make sense

### Logging

### Documentation

### DIFF
--- a/src/main/java/uk/gov/pay/products/service/PaymentCreator.java
+++ b/src/main/java/uk/gov/pay/products/service/PaymentCreator.java
@@ -32,6 +32,7 @@ import static net.logstash.logback.argument.StructuredArguments.kv;
 import static org.apache.commons.lang3.StringUtils.isEmpty;
 import static uk.gov.pay.products.util.RandomIdGenerator.randomUserFriendlyReference;
 import static uk.gov.pay.products.util.RandomIdGenerator.randomUuid;
+import static uk.gov.service.payments.logging.LoggingKeys.GATEWAY_ACCOUNT_ID;
 import static uk.gov.service.payments.logging.LoggingKeys.PAYMENT_EXTERNAL_ID;
 
 public class PaymentCreator {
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/products/service/PaymentCreator.java
```diff
@@ -145,7 +146,10 @@ private PaymentEntity mergePaymentEntityWithReferenceNumberCheck(PaymentEntity p
                 
                 if (PanDetector.isSuspectedPan(paymentEntity.getReferenceNumber())) {
                     logger.warn("Suspected PAN entered by user in reference field",
-                            kv(PAYMENT_EXTERNAL_ID, paymentEntity.getGovukPaymentId()));
+                            kv(PAYMENT_EXTERNAL_ID, paymentEntity.getGovukPaymentId()),
+                            kv(GATEWAY_ACCOUNT_ID, paymentEntity.getGatewayAccountId()),
+                            kv("reference_length", PanDetector.cleanedReference(paymentEntity.getReferenceNumber()).length())
```
alexbishop1, line 151: This isn’t quite the reference length, is it? It’s the length of the reference after it’s been “cleaned”. Since cleaning involves removing dashes and spaces, it could be shorter. Maybe call this `number_of_digits` or something?

src/main/java/uk/gov/pay/products/validations/PanDetector.java
```diff
@@ -10,10 +10,12 @@ public static boolean isSuspectedPan(String reference) {
         if (reference.length() < MIN_PAN_LENGTH || reference.length() > MAX_PAN_LENGTH) {
             return false;
         }
-        var cleaned = reference
-                .replace("-", "")
-                .replace(" ", "");
+        var cleaned = cleanedReference(reference);
```
alexbishop1, line 13: Maybe this `var` should be `String` now as it’s not 100% clear from the method name what it returns.

alexbishop1 (changes requested):
Looks fine. A couple of things to consider.

alexbishop1 (approved):

PR #897, opened 2022-04-05
PP-9460 Add column amount_hint to products

stephencdaly:
This column will be used for storing a hint displayed above the amount input for a payment link where the user can choose the amount.


oswaldquek (approved):
It's been a while but I assume columns are nullable by default!

PR #898, opened 2022-04-05
PP-9640 Return amount_hint field and allow update

stephencdaly:
Return the amount_hint in all API responses that return a product.
Accept an amount_hint when creating a product and when updating a
product.


kbottla (approved):
LGTM!

PR #953, opened 2022-06-20
PP-9587 Pass on 403 status from Public api

stephencdaly:
Public API will respond with a 403 status code either if the gateway account does not have PSP credentials configured or if the gateway account has been disabled.

Pass this status code on in the response for the create payment API in products so that this status code can be used by products-ui to display a more specific error message.

### DIFF
--- a/src/main/java/uk/gov/pay/products/exception/mapper/PaymentCreationExceptionMapper.java
+++ b/src/main/java/uk/gov/pay/products/exception/mapper/PaymentCreationExceptionMapper.java
@@ -8,6 +8,9 @@ import uk.gov.pay.products.util.Errors;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.ext.ExceptionMapper;
 
+import static javax.ws.rs.core.Response.Status.FORBIDDEN;
+import static javax.ws.rs.core.Response.Status.INTERNAL_SERVER_ERROR;
+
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/pay/products/exception/mapper/PaymentCreationExceptionMapper.java
```diff
@@ -8,6 +8,9 @@
 import javax.ws.rs.core.Response;
 import javax.ws.rs.ext.ExceptionMapper;
 
+import static org.eclipse.jetty.http.HttpStatus.FORBIDDEN_403;
+import static org.eclipse.jetty.http.HttpStatus.INTERNAL_SERVER_ERROR_500;
```
alexbishop1, line 12: Think I’d prefer to get the constants from `javax.ws.rs.core.Response.Status` or similar rather than a third-party library (yes, I know Java EE is now technically a third-party library!).

alexbishop1 (changes requested):

src/main/java/uk/gov/pay/products/exception/mapper/PaymentCreationExceptionMapper.java
stephencdaly: updated

alexbishop1 (approved):

PR #993, opened 2022-08-30
PP-10019 Update hint columns to text type

stephencdaly:
Update columns amount_hint and reference_hint from type VARCHAR(255) to
text so that we can increase the number of characters allowed for these
fields.

Update .gitignore to ignore generated file dependency-reduced-pom.xml

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,4 @@ bin
 
 .DS_Store
 /pacts/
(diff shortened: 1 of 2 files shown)


alan-gds (approved):

alan-gds:
I’m confident that this change does not require further review and meets the secure coding requirements of GOV.UK Pay.

PR #1032, opened 2022-10-12
PP-10035 Updated ubuntu 18.04 to ubuntu-latest

Rkotyank:
checked for Ubuntu 18.04 images and updated to ubuntu-latest

### DIFF
--- a/.github/workflows/_run-pact-provider-tests.yml
+++ b/.github/workflows/_run-pact-provider-tests.yml
@@ -24,7 +24,8 @@ permissions:
 jobs:
   run-products-as-provider-tests:
     name: Run Products as Provider
-    runs-on: ubuntu-18.04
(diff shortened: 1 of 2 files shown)


.github/workflows/_run-pact-provider-tests.yml
```diff
@@ -24,7 +24,8 @@ permissions:
 jobs:
   run-products-as-provider-tests:
     name: Run Products as Provider
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
```
katstevens, line 27: I think an extra line has sneaked in here and in in the other file. In this case it doesn't matter too much (Github Actions yaml is relatively forgiving!) but in other cases it might result in something unexpected! Just something to watch out for in the future.

katstevens (approved):

PR #1038, opened 2022-10-18
PP-10004 M1 support for integration tests

nlsteers:
## WHAT YOU DID

- removed use of `govuk/postgres:11`, baselining on architecture agnostic `postgres:11.16`
- updated test bases and rules to enable cross architecture support
- added integration tests specific liquibase changelog (`it-migrations.xml`) to initialise the database before migrations can be run
- updated GHA to use new images


SandorArpa (approved):
LGTM

PR #1078, opened 2023-01-04
BAU Unhide pay_api_token field from OpenApi spec

stephencdaly:
This field should be included as it is returned in responses.

### DIFF
--- a/openapi/products_spec.yaml
+++ b/openapi/products_spec.yaml
@@ -674,6 +674,8 @@ components:
         name:
           type: string
           example: A name for the product
+        pay_api_token:
+          type: string
         price:
           type: integer
           format: int64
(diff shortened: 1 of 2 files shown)


openapi/products_spec.yaml
```diff
@@ -674,6 +674,8 @@ components:
         name:
           type: string
           example: A name for the product
+        pay_api_token:
```
kbottla, line 677: Is there any need to return this in the API? if not, maybe we can remove it from the response?

stephencdaly: Yeh it's used by selfservice in order to delete the API key when a payment link is deleted. Possibly we should be storing the token_link instead and deleting using that but that isn't how it currently works.
I just came across this when I was fixing some cypress stubbing and needed to use this in a stub and was a bit confused when this wasn't in the API spec.

kbottla (approved):

PR #1169, opened 2023-06-15
BAU bump `dropwizard` -> 2.1.6

nlsteers:
## WHAT YOU DID

- prefer `dropwizard-dependencies` defined versions for `junit` and `mockito`
- send test output to file over console


sfount (approved):
I think deferring to the dropwizard BOM for a project build around dropwizard makes sense, I don't have all the context on these dependency sharp edge changes but this builds and looks sensible

PR #1186, opened 2023-07-05
PP-11167 Handle card number in payment link reference error code

kbottla:
## WHAT
- Handles P0105 error code from publicAPI and includes error identifier in the response to products-ui
- Redacts reference from products DB when the error code is P0105


stephencdaly (approved):
LGTM!

PR #1242, opened 2023-10-17
PP-11652 Upgrade EclipseLink from 2.7.11 to 2.7.13

alexbishop1:
Unlike version 2.7.11, 2.7.13 works with Java 21.

with @SandorArpa


SandorArpa (approved):
I’m confident that this change does not require further review and meets the secure coding requirements of GOV.UK Pay

PR #1329, opened 2024-03-05
bau: Card number in payment link reference is not an error

oswaldquek:
A card number entered in a payment link reference is not something that should be sent to sentry.


james-peacock-gds (approved):
LGTM!

PR #1409, opened 2024-07-22
PP-12712 Log payment creation errors at WARN level

james-peacock-gds:
- If a payment link fails due to a misconfigured account, we want a log at WARN level rather than ERROR, to avoid Sentry noise
- Amend logging statements to WARN if there is an error in the response from publicapi
- A previous PR ensured that the resulting PaymentCreationException will log at WARN level if the cause is a misconfigured account (but did not address the ERROR logs in the create payment process covered here)
- PaymentCreationException will continue to log at ERROR level for other causes of payment creation error


nlsteers (approved):
💯

PR #1518, opened 2025-04-01
PP-13655 upgrade modules to dropwizard 4

SandorArpa:
## WHAT YOU DID

Upgrade to use Dropwizard 4.
Also upgrade, what we can to Java Jakarta EE.


pom.xml
```diff
@@ -278,16 +225,17 @@
...
-            <artifactId>swagger-annotations</artifactId>
+            <artifactId>swagger-core-jakarta</artifactId>
             <version>${swagger-version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.eclipse.parsson</groupId>
+            <artifactId>parsson</artifactId>
+            <version>1.1.7</version>
+            <scope>test</scope>
+        </dependency>
```
alexbishop1, line 238: What is this and why do we need it?

src/test/java/uk/gov/pay/products/matchers/PaymentEntityMatcher.java
```diff
@@ -1,7 +1,7 @@
 package uk.gov.pay.products.matchers;
 
-import org.apache.commons.lang.ObjectUtils;
-import org.apache.commons.lang.StringUtils;
+import org.apache.commons.lang3.ObjectUtils;
+import org.apache.commons.lang3.StringUtils;
```
alexbishop1, line 4: Are we confident there’s no behaviour changes between the two versions for the methods we use?

pom.xml
SandorArpa: It's used in PublicApiStub (test) by jakarta.json.JsonObjectBuilder as default implementation of JsonProvider. It does not come bundled in jakarta.json-api in case you want to use a different implementation.

src/test/java/uk/gov/pay/products/matchers/PaymentEntityMatcher.java
SandorArpa: This matcher is exclusively used in tests, and all existing tests pass without any modification so I think it's safe to assume that the desired behaviour did not change. Also, lang3 release docs does not mention any behaviour changes.

alexbishop1 (approved):

PR #1558, opened 2025-07-30
PP-14315 Order payment links by created date

ralphm10:
## WHAT YOU DID

- add created date (timestamp) to product response 

## How to test

- create at least two payment links locally 
- view the links in self-service using https://github.com/alphagov/pay-selfservice/pull/4684 
- the most recent payment link you created should be displayed at the top of the list

### DIFF
--- a/src/test/java/uk/gov/pay/products/resources/ProductResourceIT.java
+++ b/src/test/java/uk/gov/pay/products/resources/ProductResourceIT.java
@@ -2,6 +2,8 @@ package uk.gov.pay.products.resources;
 
 import io.restassured.response.ValidatableResponse;
 import org.apache.commons.lang3.RandomStringUtils;
+import org.hamcrest.Matcher;
+import org.hamcrest.core.CombinableMatcher;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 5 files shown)


src/test/java/uk/gov/pay/products/resources/ProductResourceIT.java
```diff
@@ -513,7 +515,8 @@ void findProductByGatewayAccountIdAndExternalId_shouldReturnProduct_whenFound()
                     .body(TYPE, is(product.getType().name()))
                     .body(DESCRIPTION, is(product.getDescription()))
                     .body(RETURN_URL, is(product.getReturnUrl()))
-                    .body(LANGUAGE, is("en"));
+                    .body(LANGUAGE, is("en"))
+                    .body(DATE_CREATED, is(notNullValue()));
```
nlsteers, line 519: unsure if we do this elsewhere but could you inject a fixed `Clock` into this test to be able to check a specific value here (as opposed to nullness)?

nlsteers (approved):
LGTM, one comment but can be merged as is

